PR #6, opened 2021-11-03
Add intial tests for XPCEncoder

amomchilov:
What it says on the tin.

### DIFF
--- a/Tests/SecureXPCTests/TestHelpers.swift
+++ b/Tests/SecureXPCTests/TestHelpers.swift
@@ -41,3 +41,22 @@ fileprivate func assertEqual(
 		XCTFail("\(actual) is not equal to \(expected).", file: file, line: line)
 	}
 }
+
+// MARK: XPC object factories
+
+/// Converts the provided `sourceArray` into an XPC array, by transforming its non-nil elements by the provided `transformIntoXPCObject` closure,
+/// and replacing `nil` values with the proper XPC null object.
+/// - Parameters:
+///   - sourceArray: The values to transform and pack into the XPC array
+///   - transformIntoXPCObject: The closures used to transform the non-nil source elements into XPC objects
+/// - Returns: an XPC array containing the transformed XPC objects
+func createXPCArray<T>(from sourceArray: [T?], using transformIntoXPCObject: (T) -> xpc_object_t) -> xpc_object_t {
+	let xpcArray = xpc_array_create(nil, 0)
+	
+	for element in sourceArray {
+		let xpcObject = element.map { transformIntoXPCObject($0) } ?? xpc_null_create()
+		xpc_array_append_value(xpcArray, xpcObject)
+	}
+	
+	return xpcArray
+}
--- a/Tests/SecureXPCTests/TestHelpers.swift
+++ b/Tests/SecureXPCTests/TestHelpers.swift
@@ -1,0 +1,43 @@
+//
+//  TestHelpers.swift
+//  
+//
+//  Created by Alexander Momchilov on 2021-11-03.
+//
+
+import XCTest
+@testable import SecureXPC
+
+// MARK: Encoding entry point
+
+func encode<T: Encodable>(_ input: T) throws -> xpc_object_t {
+	try XPCEncoder.encode(input)
+}
+
+// MARK: Assertions
+
+
+/// Assert that the provided `input`, when encoded using an XPCEncoder, is equal to the `expected` XPC Object
+func assert<T: Encodable>(
+	_ input: T,
+	encodesEqualTo expected: xpc_object_t,
+	file: StaticString = #file,
+	line: UInt = #line
+) throws {
+	let actual = try encode(input)
+	assertEqual(actual, expected, file: file, line: line)
+}
+
+/// Asserts that `actual` and `expected` are value-equal, according to `xpc_equal`
+///
+/// `-[OS_xpc_object isEqual]` exists, but it uses object-identity as the basis for equality, which is not what we want here.
+fileprivate func assertEqual(
+	_ actual: xpc_object_t,
+	_ expected: xpc_object_t,
+	file: StaticString = #file,
+	line: UInt = #line
+) {
+	if !xpc_equal(expected, actual) {
+		XCTFail("\(actual) is not equal to \(expected).", file: file, line: line)
+	}
+}
--- a/Tests/SecureXPCTests/TestHelpers.swift
+++ b/Tests/SecureXPCTests/TestHelpers.swift
@@ -60,3 +60,23 @@ func createXPCArray<T>(from sourceArray: [T?], using transformIntoXPCObject: (T)
 	
 	return xpcArray
 }
+
+/// Converts the provided `sourceDict` into an XPC dictionary, by transforming its non-nil values by the provided `transformIntoXPCObject` closure,
+/// and replacing `nil` values with the proper XPC null object.
+/// - Parameters:
+///   - sourceDict: The values to transform and pack into the XPC dictionary
+///   - transformIntoXPCObject: The closures used to transform the non-nil source elements into XPC objects
+/// - Returns: an XPC dictionary containing the transformed XPC objects
+func createXPCDict<V>(from sourceDict: [String: V?], using wrapIntoXPCObject: (V) -> xpc_object_t) -> xpc_object_t {
+	let xpcDict = xpc_dictionary_create(nil, nil, 0)
+	
+	for (key, value) in sourceDict {
+		let xpcValue = value.map(wrapIntoXPCObject) ?? xpc_null_create()
+		
+		key.withCString { keyP in
+			xpc_dictionary_set_value(xpcDict, keyP, xpcValue)
+		}
+	}
+	
+	return xpcDict
+}
--- a/Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
+++ b/Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
@@ -145,4 +145,18 @@ final class XPCEncoder_ArrayEncodingTests: XCTestCase {
 		
 		try assert(nestedArray, encodesEqualTo: expectedXPCArray)
 	}
+	
+	func testEncodeArrayOfDictionaries() throws {
+		// There's too many possible permutations, but it should be satisfactory to just test one kind of nesting.
+		let arrayOfDicts: [[String: Int64]] = [
+			["a": 1, "b": 2, "c": 3],
+			["e": 4, "f": 5, "g": 6],
+		]
+		
+		let expectedXPCArray = createXPCArray(from: arrayOfDicts, using: { subDict in
+			createXPCDict(from: subDict, using: xpc_int64_create)
+		})
+		
+		try assert(arrayOfDicts, encodesEqualTo: expectedXPCArray)
+	}
 }
--- a/Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
+++ b/Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
@@ -1,0 +1,148 @@
+import XCTest
+@testable import SecureXPC
+
+final class XPCEncoder_ArrayEncodingTests: XCTestCase {
+	// MARK: Signed Integers
+	
+	func testEncodes_arrayOf_SignedIntegers_asArrayOf_XPCInts() throws {
+		let   ints: [Int  ] = [.min, -123, 0, 123, .max]
+		let  int8s: [Int8 ] = [.min, -123, 0, 123, .max]
+		let int16s: [Int16] = [.min, -123, 0, 123, .max]
+		let int32s: [Int32] = [.min, -123, 0, 123, .max]
+		let int64s: [Int64] = [.min, -123, 0, 123, .max]
+		
+		try assert(  ints, encodesEqualTo: createXPCArray(from:   ints, using: { xpc_int64_create(Int64($0)) }))
+		try assert( int8s, encodesEqualTo: createXPCArray(from:  int8s, using: { xpc_int64_create(Int64($0)) }))
+		try assert(int16s, encodesEqualTo: createXPCArray(from: int16s, using: { xpc_int64_create(Int64($0)) }))
+		try assert(int32s, encodesEqualTo: createXPCArray(from: int32s, using: { xpc_int64_create(Int64($0)) }))
+		try assert(int64s, encodesEqualTo: createXPCArray(from: int64s, using: { xpc_int64_create(Int64($0)) }))
+	}
+	
+	func testEncodes_arrayOf_UnsignedIntegers_asArrayOf_XPCUInts() throws {
+		let   uints: [UInt  ] = [.min, 0, 123, .max]
+		let  uint8s: [UInt8 ] = [.min, 0, 123, .max]
+		let uint16s: [UInt16] = [.min, 0, 123, .max]
+		let uint32s: [UInt32] = [.min, 0, 123, .max]
+		let uint64s: [UInt64] = [.min, 0, 123, .max]
+		
+		try assert(  uints, encodesEqualTo: createXPCArray(from:   uints, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert( uint8s, encodesEqualTo: createXPCArray(from:  uint8s, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(uint16s, encodesEqualTo: createXPCArray(from: uint16s, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(uint32s, encodesEqualTo: createXPCArray(from: uint32s, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(uint64s, encodesEqualTo: createXPCArray(from: uint64s, using: { xpc_uint64_create(UInt64($0)) }))
+	}
+	
+	// MARK: Floating point numbers
+	
+	func testEncodes_arrayOf_Floats_asArrayOf_XPCDoubles() throws {
+		let floats: [Float] = [
+			-Float.infinity,
+			-Float.greatestFiniteMagnitude,
+			-123,
+			-0.0,
+			 0.0,
+			 123,
+			 Float.greatestFiniteMagnitude,
+			 Float.infinity
+		]
+		
+		try assert(floats, encodesEqualTo: createXPCArray(from: floats, using: { xpc_double_create(Double($0)) }))
+		
+		// These don't have regular equality, so we'll check them seperately.
+		let nans = try encode([Float.nan, Float.signalingNaN])
+		XCTAssertEqual(xpc_array_get_count(nans), 2)
+		XCTAssert(xpc_array_get_double(nans, 0).isNaN)
+		XCTAssert(xpc_array_get_double(nans, 1).isNaN)
+	}
+	
+	func testEncodes_arrayOf_Doubles_asArrayOf_XPCDoubles() throws {
+		let doubles: [Double] = [
+			-Double.infinity,
+			-Double.greatestFiniteMagnitude,
+			-123,
+			-0.0,
+			 0.0,
+			 123,
+			 Double.greatestFiniteMagnitude,
+			 Double.infinity
+		]
+		
+		try assert(doubles, encodesEqualTo: createXPCArray(from: doubles, using: { xpc_double_create(Double($0)) }))
+		
+		// These don't have regular equality, so we'll check them seperately.
+		let nans = try encode([Double.nan, Double.signalingNaN])
+		XCTAssertEqual(xpc_array_get_count(nans), 2)
+		XCTAssert(xpc_array_get_double(nans, 0).isNaN)
+		XCTAssert(xpc_array_get_double(nans, 1).isNaN)
+	}
+	
+	// MARK: Misc. types
+	
+	func testEncodes_arrayOf_Bools_asArrayOf_XPCBools() throws {
+		let bools = [false, true]
+		try assert(bools, encodesEqualTo: createXPCArray(from: bools, using: { xpc_bool_create($0) }))
+	}
+	
+	func testEncodes_arrayOf_Strings_asArrayOf_XPCStrings() throws {
+		let strings = ["", "Hello, world!"]
+		try assert(strings, encodesEqualTo: createXPCArray(from: strings, using: { str in
+			str.withCString(xpc_string_create)
+		}))
+	}
+	
+	func testEncodes_arrayOf_nils() throws {
+		// Signed integers
+		let   ints: [Optional<Int  >] = [.min, -123, 0, 123, .max, nil]
+		let  int8s: [Optional<Int8 >] = [.min, -123, 0, 123, .max, nil]
+		let int16s: [Optional<Int16>] = [.min, -123, 0, 123, .max, nil]
+		let int32s: [Optional<Int32>] = [.min, -123, 0, 123, .max, nil]
+		let int64s: [Optional<Int64>] = [.min, -123, 0, 123, .max, nil]
+		
+		try assert(  ints, encodesEqualTo: createXPCArray(from:   ints, using: { xpc_int64_create(Int64($0)) }))
+		try assert( int8s, encodesEqualTo: createXPCArray(from:  int8s, using: { xpc_int64_create(Int64($0)) }))
+		try assert(int16s, encodesEqualTo: createXPCArray(from: int16s, using: { xpc_int64_create(Int64($0)) }))
+		try assert(int32s, encodesEqualTo: createXPCArray(from: int32s, using: { xpc_int64_create(Int64($0)) }))
+		try assert(int64s, encodesEqualTo: createXPCArray(from: int64s, using: { xpc_int64_create(Int64($0)) }))
+		
+		// Unsigned integers
+		let   uints: [Optional<UInt  >] = [.min, 0, 123, .max, nil]
+		let  uint8s: [Optional<UInt8 >] = [.min, 0, 123, .max, nil]
+		let uint16s: [Optional<UInt16>] = [.min, 0, 123, .max, nil]
+		let uint32s: [Optional<UInt32>] = [.min, 0, 123, .max, nil]
+		let uint64s: [Optional<UInt64>] = [.min, 0, 123, .max, nil]
+		
+		try assert(  uints, encodesEqualTo: createXPCArray(from:   uints, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert( uint8s, encodesEqualTo: createXPCArray(from:  uint8s, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(uint16s, encodesEqualTo: createXPCArray(from: uint16s, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(uint32s, encodesEqualTo: createXPCArray(from: uint32s, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(uint64s, encodesEqualTo: createXPCArray(from: uint64s, using: { xpc_uint64_create(UInt64($0)) }))
+		
+		// Floating point numbers
+		let floats: [Float?] = [-123, 0, 123, nil]
+		try assert(floats, encodesEqualTo: createXPCArray(from: floats, using: { xpc_double_create(Double($0)) }))
+		let doubles: [Double?] = [-123, 0, 123, nil]
+		try assert(floats, encodesEqualTo: createXPCArray(from: doubles, using: { xpc_double_create($0) }))
+		
+		// Misc. types
+		let bools: [Bool?] = [false, true, nil]
+		try assert(bools, encodesEqualTo: createXPCArray(from: bools, using: { xpc_bool_create($0) }))
+		
+		let strings: [String?] = ["", "Hello, world!", nil]
+		try assert(strings, encodesEqualTo: createXPCArray(from: strings, using: { str in
+			str.withCString(xpc_string_create)
+		}))
+	}
+	
+	// MARK: Arrays of aggregates
+	
+	func testEncode_arrayOf_Arrays() throws {
+		// There's too many possible permutations, but it should be satisfactory to just test one kind of nesting.
+		let nestedArray: [[Int64]] = [[1, 2], [3, 4]]
+		
+		let expectedXPCArray = createXPCArray(from: nestedArray, using: { row in
+			createXPCArray(from: row, using: xpc_int64_create)
+		})
+		
+		try assert(nestedArray, encodesEqualTo: expectedXPCArray)
+	}
+}
--- a/Tests/SecureXPCTests/XPCEncoder/Dictionary encoding tests.swift
+++ b/Tests/SecureXPCTests/XPCEncoder/Dictionary encoding tests.swift
@@ -1,0 +1,173 @@
+//
+//  Dictinary encoding tests.swift
+//
+//
+//  Created by Alexander Momchilov on 2021-11-03.
+//
+
+import XCTest
+@testable import SecureXPC
+
+final class XPCEncoder_DictionaryEncodingTests: XCTestCase {
+	// MARK: Signed Integers
+	
+	func testEncodes_dictOf_SignedIntegers_asDictOf_XPCInts() throws {
+		let dictOfInt:   [String: Int  ] = ["int"  : 123]
+		let dictOfInt8:  [String: Int8 ] = ["int8" : 123]
+		let dictOfInt16: [String: Int16] = ["int16": 123]
+		let dictOfInt32: [String: Int32] = ["int32": 123]
+		let dictOfInt64: [String: Int64] = ["int64": 123]
+		
+		try assert(  dictOfInt, encodesEqualTo: createXPCDict(from:   dictOfInt, using: { xpc_int64_create(Int64($0)) }))
+		try assert( dictOfInt8, encodesEqualTo: createXPCDict(from:  dictOfInt8, using: { xpc_int64_create(Int64($0)) }))
+		try assert(dictOfInt16, encodesEqualTo: createXPCDict(from: dictOfInt16, using: { xpc_int64_create(Int64($0)) }))
+		try assert(dictOfInt32, encodesEqualTo: createXPCDict(from: dictOfInt32, using: { xpc_int64_create(Int64($0)) }))
+		try assert(dictOfInt64, encodesEqualTo: createXPCDict(from: dictOfInt64, using: { xpc_int64_create(Int64($0)) }))
+	}
+	
+	func testEncodes_dictOf_UnsignedIntegers_asDictOf_XPCUInts() throws {
+		let dictOfUInt:   [String: UInt  ] = ["uint"  : 123]
+		let dictOfUInt8:  [String: UInt8 ] = ["uint8" : 123]
+		let dictOfUInt16: [String: UInt16] = ["uint16": 123]
+		let dictOfUInt32: [String: UInt32] = ["uint32": 123]
+		let dictOfUInt64: [String: UInt64] = ["uint64": 123]
+		
+		try assert(  dictOfUInt, encodesEqualTo: createXPCDict(from:   dictOfUInt, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert( dictOfUInt8, encodesEqualTo: createXPCDict(from:  dictOfUInt8, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(dictOfUInt16, encodesEqualTo: createXPCDict(from: dictOfUInt16, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(dictOfUInt32, encodesEqualTo: createXPCDict(from: dictOfUInt32, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(dictOfUInt64, encodesEqualTo: createXPCDict(from: dictOfUInt64, using: { xpc_uint64_create(UInt64($0)) }))
+	}
+	
+	// MARK: Floating point numbers
+	
+	func testEncodes_dictOf_Floats_asDictOf_XPCDoubles() throws {
+		let dictOfFloats: [String: Float] = [
+			"-infinity": -.infinity,
+			"-greatestFiniteMagnitude": -.greatestFiniteMagnitude,
+			"-123": -123,
+			"-0.0": -0.0,
+			"0.0": 0.0,
+			"123": 123,
+			"greatestFiniteMagnitude": .greatestFiniteMagnitude,
+			"infinity": .infinity,
+		]
+
+		try assert(dictOfFloats, encodesEqualTo: createXPCDict(from: dictOfFloats, using: { xpc_double_create(Double($0)) }))
+
+		// These don't have regular equality, so we'll check them seperately.
+		let nans = try encode(["nan": Float.nan, "signalingNaN": Float.signalingNaN])
+		XCTAssertEqual(xpc_dictionary_get_count(nans), 2)
+		XCTAssert(xpc_dictionary_get_double(nans, "nan").isNaN)
+		XCTAssert(xpc_dictionary_get_double(nans, "signalingNaN").isNaN)
+	}
+	
+	func testEncodes_dictOf_Doubles_asDictOf_XPCDoubles() throws {
+		let dictOfDoubles: [String: Double] = [
+			"-infinity": -.infinity,
+			"-greatestFiniteMagnitude": -.greatestFiniteMagnitude,
+			"-123": -123,
+			"-0.0": -0.0,
+			"0.0": 0.0,
+			"123": 123,
+			"greatestFiniteMagnitude": .greatestFiniteMagnitude,
+			"infinity": .infinity,
+		]
+
+		try assert(dictOfDoubles, encodesEqualTo: createXPCDict(from: dictOfDoubles, using: { xpc_double_create(Double($0)) }))
+
+		// These don't have regular equality, so we'll check them seperately.
+		let nans = try encode(["nan": Double.nan, "signalingNaN": Double.signalingNaN])
+		XCTAssertEqual(xpc_dictionary_get_count(nans), 2)
+		XCTAssert(xpc_dictionary_get_double(nans, "nan").isNaN)
+		XCTAssert(xpc_dictionary_get_double(nans, "signalingNaN").isNaN)
+	}
+	
+	// MARK: Misc. types
+	
+	func testEncodes_dictOf_Bools_asDictOf_XPCBools() throws {
+		let bools: [String: Bool?] = ["false": false, "true": true, "nil": nil]
+		try assert(bools, encodesEqualTo: createXPCDict(from: bools, using: { xpc_bool_create($0) }))
+	}
+	
+	func testEncodes_dictOf_Strings_asDictOf_XPCStrings() throws {
+		let strings: [String: String] = ["empty": "", "string": "Hello, world!"]
+		try assert(strings, encodesEqualTo: createXPCDict(from: strings, using: { str in
+			str.withCString(xpc_string_create)
+		}))
+	}
+	
+	func testEncodes_dictsOf_Nils() throws {
+		// Signed integers
+		let dictOfInt:   [String: Optional<Int  >] = ["int"  : 123, "nil": nil]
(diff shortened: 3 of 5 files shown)


amomchilov:
@trilemma-dev what do you think?

jakaplan:
Thanks for putting this together. Will review; may take me a few days to get through all of it. FYI I'm in the NZDT timezone.

jakaplan:
I read through the PR and this brought up several important points that are better discussed at a high level before getting into any specifics on the details of the code:
 - `XPCEncoder` and `XPCDecoder` aren't intended to be a general purpose encoder or decoder respectively, which is in part why they're not part of the public API. They only exist for the purpose of serializing/deserializing `Codable` instances across an XPC connection.
     - So for example while XPC has a unique concept of a "data" type, `Codable` does not. The Swift type `Data` knows how encode and decode itself (I believe it chooses to use an unkeyed container with `UInt8` elements, but that's its own hidden implementation detail). This is equally the case for `UUID` and `Date`. All of these Swift types are `Codable` and therefore can be used with SecureXPC without issue and without any special casing needed.
     -  File descriptors and shared memory are intentionally *not* supported as they are concepts incompatible with `Codable` which is designed for indefinite serialization. As such choosing `Codable` for SecureXPC had tradeoffs because XPC does not have the same requirement — it can represent live resources. However, I decided the tradeoff was worth it because of Swift's excellent built in support for Codable (particularly the compiler automatically generating conformance in many cases). If there was a desire to support file descriptors and shared memory, I can envision ways to add them in the future, although it would mean expanding the API quite considerably.
 - Currently SecureXPC makes no guarantees that the internal serialization format it uses will remain stable between releases. This is something I **do** want to figure out prior to a 1.0.0 release as I believe it should be considered part of the SemVer criteria. So for example a guarantee could exist that all 1.x clients can talk to a 1.x servers, and vice versa. In its current 0.x form, that's not true.
    - The unit tests you've written are asserting specific ways of encoding Swift values into XPC values, which is *not* currently a guarantee of this framework. Once a serialization format is stable, having such a set of tests would be quite valuable to ensure that the format does not change except intentionally between major versions. 
    - What is a *guarantee* of the current API (but lacks unit tests to do so) is that a `Codable` instance sent through API _can_ be perfectly decoded upon receipt such that equality is preserved. So unit tests which invoke both the encoder and decoder together could validate this. As is quite evident I have written such unit tests, but this was what I had been loosely envisioning. **Would you be open to writing tests at this granularity of unit?**
 - Your PR made very clear that decoupling the actual encoding and decoding from the "packaging" format which includes the route information would improve testability. I can make that change relatively soon.

amomchilov:
Some thoughts, in (kind of) reverse order.

> Your PR made very clear that decoupling the actual encoding and decoding from the "packaging" format which includes the route information would improve testability. I can make that change relatively soon.

That's an untested code path, so I'm afraid to make that change. If you make the change, I suppose you use whatever  SecureXPC-using project that you have to test it :D

> What is a guarantee of the current API (but lacks unit tests to do so) is that a Codable instance sent through the API can be decoded upon receipt such that equality is preserved. So unit tests which invoke both the encoder and decoder together could validate this. As is quite evident I have not written such unit tests, but this was what I had been loosely envisioning. Would you be open to writing tests at this granularity of unit?

Yes, absolutely! :)

I suppose I should have communicated my higher level intentions for the test suite. I think it should consist of:

1. A test of the encoder, in isolation (this test). 

    > The unit tests you've written are asserting specific ways of encoding Swift values into XPC values, which is not currently a guarantee of this framework. Once a serialization format is stable, having such a set of tests would be quite valuable to ensure that the format does not change except intentionally between major versions.
    
    I think updating these tests is easy enough that it's worth checking them in early. It'll at least notify us when the serialization format changes (which is acceptable pre-1.0.0, but we'd still want to know about it).
    
2. A test of the decoder, in isolation, for the same reason.
3. A round-trip test, like what you were saying. Strictly speaking, this isn't necessary when the encoder and decoder tests are good enough and consistent with each other (the expected inputs of one are the expected outputs on the other). However, in the past I've found that adding an round-trip test suite as an extra step useful, because it guarantees that consistency rather than hoping for it.
4. An integration test that launches a test XPC service and round-trips messages through it. Though IDK how to package an XPC service in a Swift package.
5. Other stuff...

Part of my motivation here was that I wanted to locally refactor to see if I can make the `*Container` classes write directly into xpc arrays/dictionaries, without the intermediate Swift array/dicts, to see how much of a performance lift that might have.

<hr>

> So for example while XPC has a unique concept of a "data" type, Codable does not. The Swift type `Data` knows how encode and decode itself (I believe it chooses to use an unkeyed container with UInt8 elements, but that's its own hidden implementation detail). This is equally the case for `UUID` and `Date`. All of these Swift types are `Codable` and therefore can be used with SecureXPC without issue and without any special casing needed.

I think this is easy enough to do that it might be worth doing.

In particular, I've noticed that `Data` is *really* inefficient to decode. As it is today, it bloats sizes by 8x. It tries to encode itself as a `[UInt8]`, but each byte gets widened into an 8 byte `int64` in order to be passed to `xpc_int64_create`. Luckily, there is not per-number object allocation overhead there (that would have been brutal for my use-case haha), because the only possible int64 values here would be 0-255, all of which fit into the tagged pointer representation.

You can get a rough of idea of what that might look like, here: https://github.com/amomchilov/SecureXPC/compare/add-intial-tests-for-XPCEncoder...Alex/new-scalar-support . Of course, I only changed the encoder half, and the decoder would need to match.

<hr>

> File descriptors and shared memory are intentionally not supported as they are concepts incompatible with Codable which is designed for indefinite serialization and cross-device transfer. As such choosing Codable for SecureXPC had tradeoffs because XPC does not have the same requirements — it can represent live resources on a single device. However, I decided the tradeoff was worth it because of Swift's excellent built in support for Codable (particularly the compiler automatically generating conformance in many cases). If there was a desire to support file descriptors and shared memory, I can envision ways to add them in the future, although it would mean expanding the API quite considerably.

I don't have an immediate need for passing FDs or shared memory regions in my program yet, but my app *does* need to pass an `xpc_endpoint_t`. My main app is sandboxed, so the only way it can use `SMBless` is by delegating that work to an unsandboxed XPC helper service. Once the installation is done, it opens the XPC connection to the privileged helper service, and passes the endpoint back to the main app.

This is the approach that's recommended in Apple's [EvenBetterAuthorizationSample](https://developer.apple.com/library/archive/samplecode/EvenBetterAuthorizationSample/Listings/Read_Me_About_EvenBetterAuthorizationSample_txt.html#//apple_ref/doc/uid/DTS40013768-Read_Me_About_EvenBetterAuthorizationSample_txt-DontLinkElementID_17), so I think supporting `xpc_endpoint_t` is pretty crucial. So at a minimum, at least *some* level of special-casing needs to happen.

jakaplan:
> That's an untested code path, so I'm afraid to make that change. If you make the change, I suppose you use whatever SecureXPC-using project that you have to test it :D

I've made this change and committed it. It should now be trivial to directly test the encoder or decoder independent of the rest of the system.

I'll reply to the rest of your comments tomorrow as it's a bit after 2AM here (got a bit carried away making this refactor)!

amomchilov:
> got a bit carried away making this refactor)!

Haha I sure know that feeling! Good night!

amomchilov:
Dang it, I regret force-pushing that. Oh well, doesn't matter, I got rid of the `__payload` unwrapping in the `encode` test helper.

🧹 super clean!

jakaplan:
Thanks for providing the additional context. I'm very welcoming of contributions to this framework, just wanted to make sure there was a shared understanding of how and why things work in this space. In particular, I didn't want you to have a mistaken understanding that by committing this test of tests that the on wire protocol would remain stable (pre-1.0.0 that is).

> 1. A test of the encoder, in isolation (this test).

Sounds good.

> 2. A test of the decoder, in isolation, for the same reason.

Makes sense.

> 3. A round-trip test, like what you were saying. Strictly speaking, this isn't necessary when the encoder and decoder tests are good enough and consistent with each other (the expected inputs of one are the expected outputs on the other). However, in the past I've found that adding an round-trip test suite as an extra step useful, because it guarantees that consistency rather than hoping for it.

I believe this is necessary in order to test that arbitrary `Codable` instances can be properly transferred. Because it's the internal implementation detail of each `Codable` conforming type and not the SecureXPC framework, I don't think it'd be a good idea to assert the specific manner in which such types encode and decode themselves. (If we were ever to provide optimized encoding and decoding for some types, that's excluded from this statement.)

> 4. An integration test that launches a test XPC service and round-trips messages through it. Though IDK how to package an XPC service in a Swift package.

I haven't fully thought this through, so not actually sure if this will work but I think this ought to be feasible using a Launch Agent. They run as the user, not root, and so I'm hoping it's possible to package these up.

Just to be extra explicit, this framework is for XPC Mach Service communication, which is not what [XPC Services](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingXPCServices.html) use (yeah it's stupidly confusing how Apple overloaded terms here).

> 5. Other stuff...

At the very least, I want to add testing for routes. If there was ever a regression here it'd likely be very disruptive.

> Part of my motivation here was that I wanted to locally refactor to see if I can make the *Container classes write directly into xpc arrays/dictionaries, without the intermediate Swift array/dicts, to see how much of a performance lift that might have.

Intriguing. It's not obvious to me that's possible (at least in a useful way). Much of the encoding does not occur at the time a call is made to the encoder, but instead at the "end" of the process when `XPCEncoder` calls `encodedValue()` on the `XPCEncoderImpl` instance it creates. To see an example of how this works, take look at `XPCKeyedEncodingContainer`'s `superEncoder()` function which is required to return an `Encoder`. The returned encoder itself is set as the _intermediate_ value in the container's `values` dictionary and then its actual encoding is deferred until `encodedValue()` is called on the container. This process can occur arbitrarily many layers deep. It's completely possible there's a different design pattern which can be used here that I haven't considered.

---

> [M]y app does need to pass an xpc_endpoint_t. My main app is sandboxed, so the only way it can use SMBless is by delegating that work to an unsandboxed XPC helper service. Once the installation is done, it opens the XPC connection to the privileged helper service, and passes the endpoint back to the main app.

> This is the approach that's recommended in Apple's EvenBetterAuthorizationSample, so I think supporting xpc_endpoint_t is pretty crucial. So at a minimum, at least some level of special-casing needs to happen.

Oof, okay this is going to be rather complicated. Will need a bunch of thought. If you weren't already aware, the term "sandbox" used by EvenBetterAuthorizationSample does not refer to same thing as the sandboxing macOS uses today. The sample is just that old :)

---  
Will begin reviewing your PR again in light of the above. May take me a few days to get through it; depends how nice the weather here is in Auckland this weekend!

Tests/SecureXPCTests/XPCEncoder/Dictionary encoding tests.swift
```diff
@@ -0,0 +1,248 @@
...
+		try assert(floats, encodesEqualTo: createXPCDict(fromOptionalElements: floats, using: { xpc_double_create(Double($0)) }))
+		let doubles: [String: Double?] = ["double": 123, "nil": nil]
+		try assert(doubles, encodesEqualTo: createXPCDict(fromOptionalElements: doubles, using: { xpc_double_create($0) }))
+
+		// Misc. types
+		
+		let bools: [String: Bool?] = ["false": false, "true": true, "nil": nil]
+		try assert(bools, encodesEqualTo: createXPCDict(fromOptionalElements: bools, using: { xpc_bool_create($0) }))
+		
+		// 💾
```
jakaplan, line 175: Same feedback as elsewhere that this is not in scope of `XPCEncoder` behavior

Tests/SecureXPCTests/XPCEncoder/Dictionary encoding tests.swift
```diff
@@ -0,0 +1,248 @@
...
+		throw XCTSkip("Support for Data is not implemented, the contents just end up as a String.")
+		
+		let data: [String: Data?] = ["empty": Data(), "data": "Hello, world!".data(using: .utf8)!, "nil": nil]
+		try assert(data, encodesEqualTo: createXPCDict(fromOptionalElements: data, using: { datum in
+			datum.withUnsafeBytes { bufferP in xpc_data_create(bufferP.baseAddress, bufferP.count) }
+		}))
+		
+		// 📆
```
jakaplan, line 183: Same for `Date`

Tests/SecureXPCTests/XPCEncoder/Dictionary encoding tests.swift
```diff
@@ -0,0 +1,248 @@
...
+			xpc_date_create(Int64(date.timeIntervalSince1970))
+		}))
+
+		let strings: [String: String?] = ["empty": "", "string": "Hello, world!", "nil": nil]
+		try assert(strings, encodesEqualTo: createXPCDict(fromOptionalElements: strings, using: { str in
+			// Does xpc_string_create copy the string? This would be a dangling pointer otherwise.
+			str.withCString(xpc_string_create)
+		}))
+		
+		throw XCTSkip("Support for UUID is not implemented, the contents just end up as a String.")
```
jakaplan, line 197: Same for `UUID`

Tests/SecureXPCTests/XPCEncoder/Dictionary encoding tests.swift
```diff
@@ -0,0 +1,248 @@
...
+		throw XCTSkip("Support for Date is not implemented, the contents just end up as a String.")
+		
+		let dates: [String: Date?] = ["distantPast": .distantPast, "now": .init(), "distantFuture": .distantFuture, "nil": nil]
+		try assert(dates, encodesEqualTo: createXPCDict(fromOptionalElements: dates, using: { date in
+			xpc_date_create(Int64(date.timeIntervalSince1970))
+		}))
+
+		let strings: [String: String?] = ["empty": "", "string": "Hello, world!", "nil": nil]
+		try assert(strings, encodesEqualTo: createXPCDict(fromOptionalElements: strings, using: { str in
+			// Does xpc_string_create copy the string? This would be a dangling pointer otherwise.
```
jakaplan, line 193: Same as noted previously, I believe the answer is yes this creates a copy

Tests/SecureXPCTests/XPCEncoder/Dictionary encoding tests.swift
```diff
@@ -0,0 +1,248 @@
...
+		let uuids: [String: UUID?] = ["uuid": UUID(), "nil": nil]
+		try assert(uuids, encodesEqualTo: createXPCDict(fromOptionalElements: uuids, using: { uuid in
+			var uuidByteTuple = uuid.uuid
+
+			return withUnsafePointer(to: &uuidByteTuple) { uuidByteTupleP in
+				uuidByteTupleP.withMemoryRebound(to: UInt8.self, capacity: 16) { uuidP in xpc_uuid_create(uuidP) }
+			}
+		}))
+
+		// TODO:
```
jakaplan, line 208: Similar feedback as for arrays that as of now #3-8 don't have framework support, so a TODO about testing it probably doesn't belong here. FWIW I suspect that `XPCEncoder` wouldn't actually be involved in making any of these happen as I can't see how `Codable` could represent them.

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
```diff
@@ -0,0 +1,240 @@
+import XCTest
+@testable import SecureXPC
+
+final class XPCEncoder_ScalarEncodingTests: XCTestCase {
+	// MARK: TODO:
```
jakaplan, line 5: Similar feedback as elsewhere that the framework doesn't support this today and that `XPCEncoder` would be unlikely to be the way that any of these have support added.

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
```diff
@@ -0,0 +1,240 @@
...
+	func testEncodes_Int32_as_XPCInt64() throws {
+		try assert(.min as Int32, encodesEqualTo: xpc_int64_create(Int64(Int32.min)))
+		try assert(-123 as Int32, encodesEqualTo: xpc_int64_create(            -123))
+		try assert(   0 as Int32, encodesEqualTo: xpc_int64_create(               0))
+		try assert( 123 as Int32, encodesEqualTo: xpc_int64_create(             123))
+		try assert(.max as Int32, encodesEqualTo: xpc_int64_create(Int64(Int32.max)))
+	}
+	
+	func testEncodes_Int64_as_XPCInt64() throws {
+		try assert(.min as Int64, encodesEqualTo: xpc_int64_create(Int64(Int64.min)))
```
jakaplan, line 47: This additional nesting inside of `Int64(...)` shouldn't be necessary

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
```diff
@@ -0,0 +1,240 @@
...
+		try assert(-123 as Int64, encodesEqualTo: xpc_int64_create(            -123))
+		try assert(   0 as Int64, encodesEqualTo: xpc_int64_create(               0))
+		try assert( 123 as Int64, encodesEqualTo: xpc_int64_create(             123))
+		try assert(.max as Int64, encodesEqualTo: xpc_int64_create(Int64(Int64.max)))
```
jakaplan, line 51: Same here

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
```diff
@@ -0,0 +1,240 @@
...
+	
+	func testEncodes_UInt32_as_XPCUInt64() throws {
+		try assert(.min as UInt32, encodesEqualTo: xpc_uint64_create(UInt64(UInt32.min)))
+		try assert(   0 as UInt32, encodesEqualTo: xpc_uint64_create(                 0))
+		try assert( 123 as UInt32, encodesEqualTo: xpc_uint64_create(               123))
+		try assert(.max as UInt32, encodesEqualTo: xpc_uint64_create(UInt64(UInt32.max)))
+	}
+	
+	func testEncodes_UInt64_as_XPCUInt64() throws {
+		try assert(.min as UInt64, encodesEqualTo: xpc_uint64_create(UInt64(UInt64.min)))
```
jakaplan, line 85: Same with regards to nesting in `UInt64(...)`

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
```diff
@@ -0,0 +1,240 @@
...
+		try assert(   0 as UInt64, encodesEqualTo: xpc_uint64_create(                 0))
+		try assert( 123 as UInt64, encodesEqualTo: xpc_uint64_create(               123))
+		try assert(.max as UInt64, encodesEqualTo: xpc_uint64_create(UInt64(UInt64.max)))
```
jakaplan, line 88: And same here as well

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
```diff
@@ -0,0 +1,240 @@
...
+	}
+	
+	// MARK: Floating point numbers
+	
+	func testEncodes_Float_as_XPCDouble() throws {
+		// "Normal" values, from lowest to highest
+		try assert(-Float.infinity,
```
jakaplan, line 95: Minor, but the indention used for all of the second parameters here strikes me as rather odd as doesn't appear to be aligned with anything in particular. I'm used to aligning all of the parameters to have the same indentation. Is there a specific formatting convention you follow/are used to?

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
```diff
@@ -0,0 +1,240 @@
...
+		
+		// NaN
+		let result = xpc_double_get_value(try encode(Float.nan))
+		XCTAssert(result.isNaN)
+	}
+	
+	func testEncodes_Float_SignalingNaN_as_XPCDouble_QuietNaN() throws {
+		let result = xpc_double_get_value(try encode(Float.signalingNaN))
+		XCTAssert(result.isNaN)
+		// IDK if this is intended or acceptable, but `Double(Float.signalingNaN).isSignalingNaN` returns `false`
```
jakaplan, line 120: Huh, but it works properly when nested in an array or dictionary?

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
```diff
@@ -0,0 +1,240 @@
...
+	// MARK: Misc. types
+	
+	func testEncodes_Bool_as_XPCBool() throws {
+		try assert(false, encodesEqualTo: xpc_bool_create(false))
+		try assert(true, encodesEqualTo: xpc_bool_create(true))
+		try assert(false, encodesEqualTo: XPC_BOOL_FALSE)
+		try assert(true, encodesEqualTo: XPC_BOOL_TRUE)
+	}
+	
+	func testEncodes_Data_as_XPCData() throws { // 💾
```
jakaplan, line 163: Same comments as elsewhere about intentional lack of support

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
```diff
@@ -0,0 +1,240 @@
...
+		}
+		
+		let sampleData = "Hello, world!".data(using: .utf8)!
+		
+		try sampleData.withUnsafeBytes { buffer in
+			try assert(sampleData, encodesEqualTo: xpc_data_create(buffer.baseAddress!, buffer.count))
+		}
+	}
+	
+	func testEncodes_Date_as_XPCDate() throws { // 📆
```
jakaplan, line 179: Same for `Date`

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
```diff
@@ -0,0 +1,240 @@
...
+		try assert(now,                encodesEqualTo: xpc_date_create(Int64(               now.timeIntervalSince1970)))
+		try assert(Date.distantFuture, encodesEqualTo: xpc_date_create(Int64(Date.distantFuture.timeIntervalSince1970)))
+	}
+	
+	func testEncodes_String_as_XPCString() throws {
+		try assert("", encodesEqualTo: xpc_string_create(""))
+		try assert("Hello, world!", encodesEqualTo: xpc_string_create("Hello, world!"))
+	}
+	
+	func testEncodes_UUID_as_XPCUUID() throws {
```
jakaplan, line 195: Same here for `UUID`

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
```diff
@@ -0,0 +1,240 @@
...
+			0x76,0x9c, // -
+			0x4e,0x50, // -
+			0xa4,0x4a, // -
+			0x53,0x35,0xf5,0x84,0xaa,0x9a,
+		])
+		
+		try assert(sampleUUID, encodesEqualTo: expectedXPCUUID)
+	}
+	
+	func testEncodes_NilValues_as_XPCNull() throws {
```
jakaplan, line 212: Hmm, is there any value in doing this for all of these different types? It seems like what this is actually testing is Swift's implementation of `Optional`?

jakaplan (commented):
Thanks! This looks like a really good test of the basics.

If your intention is to use this set of tests in order to make changes to `XPCEncoder` then the remaining portion is encoding classes which need to encode their super classes. This along with nested unkeyed contains and keyed containers makes up a lot of the complexity to how `XPCEncoder` is designed. That's because they require deferred encoding of the actual values because at the time they need to return their containers or encoders what will actually need to be encoded isn't known.

See `XPCUnkeyedEncodingContainer`'s `superEncoder()` function and `XPCKeyedEncodingContainer`'s `superEncoder()` and `superEncoder(forKey key: K)` functions for the super class encoding.

amomchilov:
##### Reply to first comment

> Just to be extra explicit, this framework is for XPC Mach Service communication, which is not what XPC Services use (yeah it's stupidly confusing how Apple overloaded terms here).

Uhhhh I didn't know this. What's the difference?

> At the very least, I want to add testing for routes. If there was ever a regression here it'd likely be very disruptive.

Indeed, I was just scratching the surface

> This process can occur arbitrarily many layers deep. It's completely possible there's a different design pattern which can be used here that I haven't considered.

I didn't quite understand this, but I'll come back to it.

> If you weren't already aware, the term "sandbox" used by EvenBetterAuthorizationSample does not refer to same thing as the sandboxing macOS uses today. The sample is just that old :)

I wasn't aware. Could you link to something that explains the difference?

> then the remaining portion is encoding classes which need to encode their super classes.

Got any ideas on how to test that? Just make a sample class hierarchy, and encode/decode it?

> See XPCUnkeyedEncodingContainer's superEncoder() function and XPCKeyedEncodingContainer's superEncoder() and superEncoder(forKey key: K) functions for the super class encoding.

Will-do!

Tests/SecureXPCTests/TestHelpers.swift
```diff
@@ -0,0 +1,117 @@
...
+//
+//  Created by Alexander Momchilov on 2021-11-03.
+//
+
+import XCTest
+@testable import SecureXPC
+
+// MARK: Encoding entry point
+
+func encode<T: Encodable>(_ input: T) throws -> xpc_object_t {
```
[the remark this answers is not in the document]
amomchilov: It used to have more purpose when it was used to wrap/unwrap the payload dicts. Otherwise it's just handy to shorted the test cases, which are already quite long.

Tests/SecureXPCTests/TestHelpers.swift
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// MARK: XPC object factories
+
+/// Converts the provided `sourceArray` into an XPC array, by transforming its elements by the provided `transformIntoXPCObject` closure.
+/// - Parameters:
+///   - sourceArray: The values to wrap and pack into the XPC array
+///   - transformIntoXPCObject: The closures used to transform the source elements into XPC objects
+/// - Returns: an XPC array containing the transformed XPC objects
+func createXPCArray<T>(from sourceArray: [T], using transformIntoXPCObject: (T) -> xpc_object_t) -> xpc_object_t {
```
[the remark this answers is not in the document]
amomchilov: There was a reason I did this, but I don't remember it, haha. I'll look into it.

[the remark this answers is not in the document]
amomchilov: I'm going to dedupe all comments. When we decide on an outcome from the main thread, i'll follow up and fix them all

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
```diff
@@ -0,0 +1,232 @@
...
+		let  int8s: [Int8 ] = [.min, -123, 0, 123, .max]
+		let int16s: [Int16] = [.min, -123, 0, 123, .max]
+		let int32s: [Int32] = [.min, -123, 0, 123, .max]
+		let int64s: [Int64] = [.min, -123, 0, 123, .max]
+		
+		try assert(  ints, encodesEqualTo: createXPCArray(from:   ints, using: { xpc_int64_create(Int64($0)) }))
+		try assert( int8s, encodesEqualTo: createXPCArray(from:  int8s, using: { xpc_int64_create(Int64($0)) }))
+		try assert(int16s, encodesEqualTo: createXPCArray(from: int16s, using: { xpc_int64_create(Int64($0)) }))
+		try assert(int32s, encodesEqualTo: createXPCArray(from: int32s, using: { xpc_int64_create(Int64($0)) }))
+		try assert(int64s, encodesEqualTo: createXPCArray(from: int64s, using: { xpc_int64_create(Int64($0)) }))
```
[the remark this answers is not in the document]
amomchilov: This is code-genned code. I didn't add a special case for `Int64`.

At worst case, this will be a no-op integer conversion. I'd be surprised if the compiler couldn't optimize it, and even if it doesn't, it wouldn't really make much of an impact on performance anyway.

However, I think it's kind of nice to keep it. I wrote these test cases in a style that tries to optimize for visual conformity, so that I can leverage the human proclivity to visually detect outliers, to detect unintentional/bad differences

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
```diff
@@ -0,0 +1,232 @@
...
+		try assert( uint8s, encodesEqualTo: createXPCArray(from:  uint8s, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(uint16s, encodesEqualTo: createXPCArray(from: uint16s, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(uint32s, encodesEqualTo: createXPCArray(from: uint32s, using: { xpc_uint64_create(UInt64($0)) }))
+		try assert(uint64s, encodesEqualTo: createXPCArray(from: uint64s, using: { xpc_uint64_create(UInt64($0)) }))
+	}
+	
+	// MARK: Floating point numbers
+	
+	func testEncodes_arrayOf_Floats_asArrayOf_XPCDoubles() throws {
+		let floats: [Float] = [
```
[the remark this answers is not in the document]
amomchilov: Oo I didn't think of those. Good call.

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
```diff
@@ -0,0 +1,232 @@
...
+	}
+	
+	// MARK: Misc. types
+	
+	func testEncodes_arrayOf_Bools_asArrayOf_XPCBools() throws {
+		let bools = [false, true]
+		try assert(bools, encodesEqualTo: createXPCArray(from: bools, using: { xpc_bool_create($0) }))
+	}
+	
+	func testEncodes_arrayOf_Data_asArrayOf_XPCData() throws { // 💾
```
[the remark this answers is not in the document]
amomchilov: > (That said, does it actually end up as a string?).


It's *much* worse than that. Storing it as a string (with say, Base64 for example), would have a 1.6x overhead (5 bits of payload per 8 bit character). The current encoding has an ***8x*** overhead. It encodes one `UInt8` of payload per `UInt64` of payload.

I think we should support the native `xpc_data` type

I think you missed a section of [one of my long comments](https://github.com/trilemma-dev/SecureXPC/pull/6#issuecomment-960865476). I quoted it below

> I think this is easy enough to do that it might be worth doing.
> 
> In particular, I've noticed that `Data` is ***really*** inefficient to encode. As it is today, it bloats sizes by 8x. It tries to encode itself as a `[UInt8]`, but each byte gets widened into an 8 byte `int64` in order to be passed to `xpc_int64_create`. Luckily, there is not per-number object allocation overhead there (that would have been brutal for my use-case haha), because the only possible int64 values here would be 0-255, all of which fit into the tagged pointer representation.
> 
> You can get a rough of idea of what that might look like, here: https://github.com/amomchilov/SecureXPC/compare/add-intial-tests-for-XPCEncoder...Alex/new-scalar-support . Of course, I only changed the encoder half, and the decoder would need to match.

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
```diff
@@ -0,0 +1,232 @@
...
+		
+		try assert(dates, encodesEqualTo: createXPCArray(from: dates, using: { date in
+			xpc_date_create(Int64(date.timeIntervalSince1970))
+		}))
+	}
+	
+	func testEncodes_arrayOf_Strings_asArrayOf_XPCStrings() throws {
+		let strings = ["", "Hello, world!"]
+		try assert(strings, encodesEqualTo: createXPCArray(from: strings, using: { str in
+			// Does xpc_string_create copy the string? This would be a dangling pointer otherwise.
```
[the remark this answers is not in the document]
amomchilov: > `stringPointer` and its `baseAddress` are only valid within that closure, so if `xpc_string_create` didn't copy the value I can't see how this would ever work.

It's not guaranteed to disappear. The standard library doesn't do anything intentionally to make the pointers become dangling. If it works on a test suite like this, that's not enough reason to be confident (low memory use, no VM paging, not much reason for memory pages to be cleared). But the fact that it works in your project is a decent indication.

To be sure, I made a quick little test. I can confirm it *does* use a copy. I'll update the comments.

```objective-c
#import <Foundation/Foundation.h>
#import <xpc/xpc.h>

int main(int argc, char *argv[]) {
	@autoreleasepool {
		char *s = strdup("Hello, world!");
		
		xpc_object_t xpc_string = xpc_string_create(s);
		
		NSLog(@"Before mutation: %@", xpc_string);
		s[0] = 'X';
		NSLog(@"After  mutation: %@", xpc_string);
		NSLog(@"New XPC  string: %@", xpc_string_create(s));
		
		/* Prints:
		2021-11-05 13:27:07.890 Untitled 2[45759:3201545] Before mutation: <OS_xpc_string: <string: 0x6000025ecb70> { length = 13, contents = "Hello, world!" }>
		2021-11-05 13:27:07.890 Untitled 2[45759:3201545] After  mutation: <OS_xpc_string: <string: 0x6000025ecb70> { length = 13, contents = "Hello, world!" }>
		2021-11-05 13:27:07.890 Untitled 2[45759:3201545] New XPC  string: <OS_xpc_string: <string: 0x6000025ecba0> { length = 13, contents = "Xello, world!" }>
		*/
	}
}
```

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
```diff
@@ -0,0 +1,232 @@
...
+			
+			return withUnsafePointer(to: &uuidByteTuple) { uuidByteTupleP in
+				uuidByteTupleP.withMemoryRebound(to: UInt8.self, capacity: 16) { uuidP in xpc_uuid_create(uuidP) }
+			}
+		}))
+	}
+	
+	func testEncodes_arrayOf_nils() throws {
+		// Signed integers
+		let   ints: [Optional<Int  >] = [.min, -123, 0, 123, .max, nil]
```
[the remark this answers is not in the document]
amomchilov: No, I wouldn't never typically write it like this, but in this case, it fits this tabular format a bit better, IMO (because you can't have a space, like `Int ?`). Compare:

```swift
let   ints: [Int?  ] = [.min, -123, 0, 123, .max, nil]
let  int8s: [Int8? ] = [.min, -123, 0, 123, .max, nil]
let int16s: [Int16?] = [.min, -123, 0, 123, .max, nil]
let int32s: [Int32?] = [.min, -123, 0, 123, .max, nil]
let int64s: [Int64?] = [.min, -123, 0, 123, .max, nil]
```

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
```diff
@@ -0,0 +1,232 @@
...
+		let uuids: [UUID?] = [UUID(), nil]
+		try assert(uuids, encodesEqualTo: createXPCArray(fromOptionalElements: uuids, using: { uuid in
+			var uuidByteTuple = uuid.uuid
+			
+			return withUnsafePointer(to: &uuidByteTuple) { uuidByteTupleP in
+				uuidByteTupleP.withMemoryRebound(to: UInt8.self, capacity: 16) { uuidP in xpc_uuid_create(uuidP) }
+			}
+		}))
+		
+		// TODO:
```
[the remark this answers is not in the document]
amomchilov: I was writing these for myself. I'll move them to an issue before shipping this PR.

> In regards to other potential TODOs would be arrays and/or dictionaries with nils and mixed types. For example an array with Bool, String, and Int32 types along with nil.

How would you even have an array of mixed types? You can't use this encoder with `NSArray`, `[Any]`, `[Encodable]` (not legal), and there's no "XPCEncodable" protocol that would let you write something like `[XPCEncodable]`

Tests/SecureXPCTests/XPCEncoder/Dictionary encoding tests.swift
amomchilov: A simple property wrapper would do. I'll make an issue and we can discuss it there

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
amomchilov: I think it started out with Xcode aligning it like 

```
assert(foo
       bar: ...)
```

And then I messed it up from there. I'll lint it all before merging.

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
amomchilov: 🤷

Tests/SecureXPCTests/XPCEncoder/Scalar encoding tests.swift
amomchilov: In effect; kind of. But that's not something that I knew from black box testing this (I wrote as much of this as possible without biasing myself by looking at the implementation)

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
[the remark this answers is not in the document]
jakaplan: Understood. If in the future custom encoding/decoding for `Data` is added then this test would make sense, but if checked in right now I think it'd be more confusing than helpful. (And unless we want to have custom support for `Date` and `UUID` cases I'd think skipped tests for those would be equally confusing.)

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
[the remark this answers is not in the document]
jakaplan: Thanks for explicitly testing that. I didn't realize the pointer _could_ be valid outside of the scope it was provided in.

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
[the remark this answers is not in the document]
jakaplan: Ah, I see. Can't say I have a strong opinion on this, was just unused to seeing it written as such.

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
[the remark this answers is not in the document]
jakaplan: Ah that's a good point. I was thinking of it from the other way around which is that an XPC array itself can be of mixed types. Getting something which encodes itself in such a manner should be feasible, but would involve a custom data structure - it wouldn't be a Swift array.

jakaplan:
> ##### Reply to first comment
> > Just to be extra explicit, this framework is for XPC Mach Service communication, which is not what XPC Services use (yeah it's stupidly confusing how Apple overloaded terms here).
> 
> Uhhhh I didn't know this. What's the difference?

XPC Services use "plain" XPC for lack of a better term. In the C API, this corresponds to an app using the [`xpc_connection_create(_:_:)`](https://developer.apple.com/documentation/xpc/1448791-xpc_connection_create) function to initiate a connection. When this is called `launchd` searches _within_ the calling app's bundle for a service bundle whose `CFBundleIdentifier` value matches the name specified in the `xpc_connection_create(_:_:)` call, then launches that XPC service (if it's not already running).

The way this communication works means it's inherently secure because all communication is scoped by `launchd` to bundles within the outermost app bundle. To quote [Apple directly](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingXPCServices.html), "Further, an XPC service is private, and is available only to the main application that contains it."

In contrast XPC for a Mach service uses the [`xpc_connection_create_mach_service(_:_:_:)`](https://developer.apple.com/documentation/xpc/1448783-xpc_connection_create_mach_servi) function. Mach services can be communicated with by many processes across the system including arbitrary apps running as standard users (although there are restrictions placed on sandboxed apps). This communication by default has no security at all and has led to numerous CVEs opened against Mach services which run as root (here's a [detailed write up about one](https://blog.obdev.at/what-we-have-learned-from-a-vulnerability/)). LaunchDaemons also have this issue and to a lesser extent LaunchAgents may as well (although they don't have a privileged escalation risk).

This is SecureXPC's reason for existing — to make it dead simple to secure these Mach service connections. Hence the intro part of the README calling out, "This framework is ideal for communicating with helper tools installed via SMJobBless."

> > This process can occur arbitrarily many layers deep. It's completely possible there's a different design pattern which can be used here that I haven't considered.
> 
> I didn't quite understand this, but I'll come back to it.

No worries, happy to discuss in-depth in the Discussions section.

> > If you weren't already aware, the term "sandbox" used by EvenBetterAuthorizationSample does not refer to same thing as the sandboxing macOS uses today. The sample is just that old :)
> 
> I wasn't aware. Could you link to something that explains the difference?

Apologies, what I wrote was misleading bordering on just incorrect. The sandbox is the same, but what's new since then is non-Mac App Store apps need to use the hardened runtime (in order to be notarized) and that interacts with the sandbox. By default this results in inheriting entitlements. This Apple Developer [thread](https://developer.apple.com/forums/thread/120647) explains what's going on. I'm no expert in this space, but happy to discuss further in the Discussions section.

> > then the remaining portion is encoding classes which need to encode their super classes.
> 
> Got any ideas on how to test that? Just make a sample class hierarchy, and encode/decode it?

Yeah so long as the super class(es) encode (at least some of) their own properties that should do it.

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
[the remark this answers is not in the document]
amomchilov: Oh yeah, it super can, it’s a typical use-after-free. You can catch it with malloc scribbling enabled, which will take the time to write over memory on free, so dangling pointers are less likely to “work” by fluke.

with move-only data types being added  to swift, this should improve. It will be possible to annotate this function in a way that ensures the pointer can’t be smuggled out

amomchilov:
> This is SecureXPC's reason for existing — to make it dead simple to secure these Mach service connections. Hence the intro part of the README calling out, "This framework is ideal for communicating with helper tools installed via SMJobBless."

But as far as I can tell, nothing stops you from using the major parts (defining routes, serializing values) with non-Mach XPC services, right? This library just doesn't have a wrapper around `xpc_connection_create` (yet?).

My main draw to it was the Codable support, removing my need to rewrite all my structs into NSObject subclasses haha.

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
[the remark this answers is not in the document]
amomchilov: Okay, I'll remove these tests and add them in a future PR where they're supported

amomchilov:
I've followed up on all the PR comments and made the appropriate changes.

* I've fixed the formatting in some places
* I've removed the comments about the string copying; we know it's safe for sure now
* I've added `leastNonZero` and `leastNormal` magnitudes in the float tests
* I've removed Data, Date and UUID related tests. I'll add them back in future PRs.

 I haven't added any tests for structs/classes. I would like to follow up on that and add it into a new PR.
 
 How do you feel about merging this work as it is, so far?

Tests/SecureXPCTests/XPCEncoder/Array encoding tests.swift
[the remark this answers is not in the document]
amomchilov: I opened an issue to track this: https://github.com/trilemma-dev/SecureXPC/issues/7

jakaplan:
> But as far as I can tell, nothing stops you from using the major parts (defining routes, serializing values) with non-Mach XPC services, right? This library just doesn't have a wrapper around xpc_connection_create (yet?).

> My main draw to it was the Codable support, removing my need to rewrite all my structs into NSObject subclasses haha.

I'm glad you've found the `Codable` support appealing, it was actually a pretty late addition (although done before I split it out into its own true framework and open sourced it on Github). Not sure if this makes sense, but `Codable` wasn't the motivating reason behind this framework (if it was I probably would've called it something like CodableXPC); validating the security of the connection was.

As far as I know there's nothing stopping the use of serialization and routes with "plain" XPC communication, although I've never tried to use it that way. Overall I imagine it's net simpler: no security considerations and the on wire protocol should be able to change without being a breaking change because an XPC Service always runs from inside the app bundle itself. There are some changes I've been considering for routes which don't make sense for XPC Services (they have to do with versioning), but in all likelihood they could be done in an optional way and just not applicable. To the extent there's some complexity here it's about how to keep the API surface simple and minimize developers using the wrong type of connection by mistake.

jakaplan (approved):
Looks good, thanks for writing this and incorporating feedback!

PR #10, opened 2021-11-08
Support for Regular XPC services

amomchilov:
IDK what the correct nomenclature is, but PR adds support for "regular" (non-Mach) XPC services.

Works great, locally.

~I extracted out `XPCBase(Client|Server)` classes to do most of the work. Perhaps it should be a protocol instead. Though I think it's correct for the client and server types to remain classes, even if they don't end up using inheritance.~

Implemented the design @jakaplan described here: https://github.com/trilemma-dev/SecureXPC/pull/10#pullrequestreview-801087521

WDYT? @jakaplan

TODOs before merging

- [ ] Update docs

### DIFF
--- a/Sources/SecureXPC/Client/XPCMachClient.swift
+++ b/Sources/SecureXPC/Client/XPCMachClient.swift
@@ -1,0 +1,26 @@
+//
+//  XPCMachClient.swift
+//  SecureXPC
+//
+//  Created by Josh Kaplan on 2021-10-09
+//
+
+import Foundation
+
+internal class XPCMachClient: XPCClient {
+	/// Creates and returns a connection for the Mach service represented by this client.
+	/// 
+	/// - Note: This client implementation intentionally does not store a reference to the ``xpc_connection_t`` as it can become
+	/// invalid for numerous reasons. Since it's expected relatively few messages will be sent and the lowest possible
+	/// latency isn't needed, it's simpler to always create the connection on demand each time a message is to be sent.
+	internal override func createConnection() -> xpc_connection_t {
+		let connection = xpc_connection_create_mach_service(self.serviceName, nil, 0)
+		xpc_connection_set_event_handler(connection, { (event: xpc_object_t) in
+			// A block *must* be set as the handler, even though this block does nothing.
+			// If it were not set, a crash would occur upon calling xpc_connection_resume.
+		})
+		xpc_connection_resume(connection)
+
+		return connection
+	}
+}
--- a/Sources/SecureXPC/Server/XPCMachServer.swift
+++ b/Sources/SecureXPC/Server/XPCMachServer.swift
@@ -1,0 +1,208 @@
+//
+//  XPCMachServer.swift
+//  
+//
+//  Created by Alexander Momchilov on 2021-11-08.
+//
+
+import Foundation
+
+public class XPCMachServer: XPCServer {
+	private let machService: xpc_connection_t
+	private let clientRequirements: [SecRequirement]
+
+	/// Creates a server that accepts requests from clients which meet the security requirements.
+	///
+	/// Because many processes on the system can talk to an XPC Mach Service, when creating a server it is required that you specifiy the security requirements
+	/// of any connecting clients:
+	/// ```swift
+	/// let reqString = """identifier "com.example.AuthorizedClient" and certificate leaf[subject.OU] = "4L0ZG128MM" """
+	/// var requirement: SecRequirement?
+	/// if SecRequirementCreateWithString(reqString as CFString,
+	///                                   SecCSFlags(),
+	///                                   &requirement) == errSecSuccess,
+	///   let requirement = requirement {
+	///    let server = XPCMachServer(machServiceName: "com.example.service",
+	///                               clientRequirements: [requirement])
+	///
+	///    <# configure and start server #>
+	/// }
+	/// ```
+	///
+	/// > Important: No requests will be processed until ``start()`` is called.
+	///
+	/// - Parameters:
+	///   - machServiceName: The name of the mach service this server should bind to. This name must be present in this program's launchd property list's
+	///                      `MachServices` entry.
+	///   - clientRequirements: If a request is received from a client, it will only be processed if it meets one (or more) of these security requirements.
+	public init(machServiceName: String, clientRequirements: [SecRequirement]) {
+		self.clientRequirements = clientRequirements
+
+		self.machService = machServiceName.withCString { serviceNamePointer in
+			return xpc_connection_create_mach_service(
+				serviceNamePointer,
+				nil, // targetq: DispatchQueue, defaults to using DISPATCH_TARGET_QUEUE_DEFAULT
+				UInt64(XPC_CONNECTION_MACH_SERVICE_LISTENER))
+		}
+	}
+
+	/// Initializes a server for a helper tool that meets
+	/// [`SMJobBless`](https://developer.apple.com/documentation/servicemanagement/1431078-smjobbless) requirements.
+	///
+	/// To successfully call this function the following requirements must be met:
+	///   - The launchd property list embedded in this helper tool must have exactly one entry for its `MachServices` dictionary
+	///   - The info property list embedded in this helper tool must have at least one element in its
+	///   [`SMAuthorizedClients`](https://developer.apple.com/documentation/bundleresources/information_property_list/smauthorizedclients)
+	///   array
+	///   - Every element in the `SMAuthorizedClients` array must be a valid security requirement
+	///     - To be valid, it must be creatable by
+	///     [`SecRequirementCreateWithString`](https://developer.apple.com/documentation/security/1394522-secrequirementcreatewithstring)
+	///
+	/// Incoming requests will be accepted from clients that meet _any_ of the `SMAuthorizedClients` requirements.
+	///
+	/// > Important: No requests will be processed until ``start()`` is called.
+	///
+	/// - Returns: A server instance initialized with the embedded property list entries.
+	public static func forBlessedHelperTool() throws -> XPCMachServer {
+		// Determine mach service name launchd property list's MachServices
+		var machServiceName: String
+		let launchdData = try readEmbeddedPropertyList(sectionName: "__launchd_plist")
+		let launchdPropertyList = try PropertyListSerialization.propertyList(
+			from: launchdData,
+			options: .mutableContainersAndLeaves,
+			format: nil) as? NSDictionary
+		if let machServices = launchdPropertyList?["MachServices"] as? [String : Any] {
+			if machServices.count == 1, let name = machServices.first?.key {
+				machServiceName = name
+			} else {
+				throw XPCError.misconfiguredBlessedHelperTool("MachServices dictionary does not have exactly one entry")
+			}
+		} else {
+			throw XPCError.misconfiguredBlessedHelperTool("launchd property list missing MachServices key")
+		}
+
+		// Generate client requirements from info property list's SMAuthorizedClients
+		var clientRequirements = [SecRequirement]()
+		let infoData = try readEmbeddedPropertyList(sectionName: "__info_plist")
+		let infoPropertyList = try PropertyListSerialization.propertyList(
+			from: infoData,
+			options: .mutableContainersAndLeaves,
+			format: nil) as? NSDictionary
+		if let authorizedClients = infoPropertyList?["SMAuthorizedClients"] as? [String] {
+			for client in authorizedClients {
+				var requirement: SecRequirement?
+				if SecRequirementCreateWithString(client as CFString, SecCSFlags(), &requirement) == errSecSuccess,
+				   let requirement = requirement {
+					clientRequirements.append(requirement)
+				} else {
+					throw XPCError.misconfiguredBlessedHelperTool("Invalid SMAuthorizedClients requirement: \(client)")
+				}
+			}
+		} else {
+			throw XPCError.misconfiguredBlessedHelperTool("Info property list missing SMAuthorizedClients key")
+		}
+		if clientRequirements.isEmpty {
+			throw XPCError.misconfiguredBlessedHelperTool("No requirements were generated from SMAuthorizedClients")
+		}
+
+		return XPCMachServer(machServiceName: machServiceName, clientRequirements: clientRequirements)
+	}
+
+	/// Read the property list embedded within this helper tool.
+	///
+	/// - Returns: The property list as data.
+	private static func readEmbeddedPropertyList(sectionName: String) throws -> Data {
+		// By passing in nil, this returns a handle for the dynamic shared object (shared library) for this helper tool
+		if let handle = dlopen(nil, RTLD_LAZY) {
+			defer { dlclose(handle) }
+
+			if let mhExecutePointer = dlsym(handle, MH_EXECUTE_SYM) {
+				let mhExecuteBoundPointer = mhExecutePointer.assumingMemoryBound(to: mach_header_64.self)
+
+				var size = UInt(0)
+				if let section = getsectiondata(mhExecuteBoundPointer, "__TEXT", sectionName, &size) {
+					return Data(bytes: section, count: Int(size))
+				} else { // No section found with the name corresponding to the property list
+					throw XPCError.misconfiguredBlessedHelperTool("Missing property list section \(sectionName)")
+				}
+			} else { // Can't get pointer to MH_EXECUTE_SYM
+				throw XPCError.misconfiguredBlessedHelperTool("Could not read property list (nil symbol pointer)")
+			}
+		} else { // Can't open handle
+			throw XPCError.misconfiguredBlessedHelperTool("Could not read property list (handle not openable)")
+		}
+	}
+
+	public override func start() -> Never {
+		// Start listener for the mach service, all received events should be for incoming connections
+		 xpc_connection_set_event_handler(self.machService, { connection in
+			 // Listen for events (messages or errors) coming from this connection
+			 xpc_connection_set_event_handler(connection, { event in
+				 self.handleEvent(connection: connection, event: event)
+			 })
+			 xpc_connection_resume(connection)
+		 })
+		 xpc_connection_resume(self.machService)
+
+		 dispatchMain()
+	}
+
+	internal override func acceptMessage(connection: xpc_connection_t, message: xpc_object_t) -> Bool {
+		// Get the code representing the client
+		var code: SecCode?
+		if #available(macOS 11, *) { // publicly documented, but only available since macOS 11
+			SecCodeCreateWithXPCMessage(message, SecCSFlags(), &code)
+		} else { // private undocumented function: xpc_connection_get_audit_token, available on prior versions of macOS
+			if var auditToken = xpc_connection_get_audit_token(connection) {
+				let tokenData = NSData(bytes: &auditToken, length: MemoryLayout.size(ofValue: auditToken))
+				let attributes = [kSecGuestAttributeAudit : tokenData] as NSDictionary
+				SecCodeCopyGuestWithAttributes(nil, attributes, SecCSFlags(), &code)
+			}
+		}
+
+		// Accept message if code is valid and meets any of the client requirements
+		var accept = false
+		if let code = code {
+			for requirement in self.clientRequirements {
+				if SecCodeCheckValidity(code, SecCSFlags(), requirement) == errSecSuccess {
+					accept = true
+				}
+			}
+		}
+
+		return accept
+	}
+
+	/// Wrapper around the private undocumented function `void xpc_connection_get_audit_token(xpc_connection_t, audit_token_t *)`.
+	///
+	/// The private undocumented function will attempt to be dynamically loaded and then invoked. If no function exists with this name `nil` will be returned. If
+	/// the function does exist, but does not match the expected signature, the process calling this function is expected to crash. However, because this is only
+	/// called on older versions of macOS which are expected to have a stable non-changing API this is very unlikely to occur.
+	///
+	/// - Parameters:
+	///   - _:  The connection for which the audit token will be retrieved for.
+	/// - Returns: The audit token or `nil` if the function could not be called.
+	private func xpc_connection_get_audit_token(_ connection: xpc_connection_t) -> audit_token_t? {
+		typealias functionSignature = @convention(c) (xpc_connection_t, UnsafeMutablePointer<audit_token_t>) -> Void
+		let auditToken: audit_token_t?
+
+		// Attempt to dynamically load the function
+		if let handle = dlopen(nil, RTLD_LAZY) {
+			defer { dlclose(handle) }
+			if let sym = dlsym(handle, "xpc_connection_get_audit_token") {
+				let function = unsafeBitCast(sym, to: functionSignature.self)
+
+				// Call the function
+				var token = audit_token_t()
+				function(connection, &token)
+				auditToken = token
+			} else {
+				auditToken = nil
+			}
+		} else {
+			auditToken = nil
+		}
+
+		return auditToken
+	}
+}
--- a/Sources/SecureXPC/Server/XPCMachServer.swift
+++ b/Sources/SecureXPC/Server/XPCMachServer.swift
@@ -7,35 +7,11 @@
 
 import Foundation
 
-public class XPCMachServer: XPCServer {
+internal class XPCMachServer: XPCServer {
 	private let machService: xpc_connection_t
 	private let clientRequirements: [SecRequirement]
 
-	/// Creates a server that accepts requests from clients which meet the security requirements.
-	///
-	/// Because many processes on the system can talk to an XPC Mach Service, when creating a server it is required that you specifiy the security requirements
-	/// of any connecting clients:
-	/// ```swift
-	/// let reqString = """identifier "com.example.AuthorizedClient" and certificate leaf[subject.OU] = "4L0ZG128MM" """
-	/// var requirement: SecRequirement?
-	/// if SecRequirementCreateWithString(reqString as CFString,
-	///                                   SecCSFlags(),
-	///                                   &requirement) == errSecSuccess,
-	///   let requirement = requirement {
-	///    let server = XPCMachServer(machServiceName: "com.example.service",
-	///                               clientRequirements: [requirement])
-	///
-	///    <# configure and start server #>
-	/// }
-	/// ```
-	///
-	/// > Important: No requests will be processed until ``start()`` is called.
-	///
-	/// - Parameters:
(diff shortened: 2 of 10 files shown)


Sources/SecureXPC/XPCBaseClient.swift
```diff
@@ -0,0 +1,113 @@
...
+    // This client implementation intentionally does not store a reference to the xpc_connection_t as it can become
+    // invalid for numerous reasons. Since it's expected relatively few messages will be sent and the lowest possible
+    // latency isn't needed, it's simpler to always create the connection on demand each time a message is to be sent.
+    
+    internal let serviceName: String
+    
+    /// Creates a client which will attempt to send messages to the specified mach service.
+    ///
+    /// - Parameters:
+    ///   - machServiceName: The name of the XPC mach service; no validation is performed on this.
```
amomchilov, line 21:
```suggestion
    ///   - serviceName: The name of the XPC service; no validation is performed on this.
```

Sources/SecureXPC/XPCBaseClient.swift
```diff
@@ -0,0 +1,113 @@
...
+            } else if xpc_equal(xpcResponse, XPC_ERROR_TERMINATION_IMMINENT) {
+                result = Result.failure(.terminationImminent)
+            } else { // Unexpected
+                result = Result.failure(.unknown)
+            }
+            reply(result)
+        })
+    }
+    
+    /// Creates and returns connection for the mach service stored by this instance of the client.
```
amomchilov, line 109:
```suggestion
    /// Creates and returns a connection for the service stored by this client.
```

Sources/SecureXPC/XPCMachClient.swift
```diff
@@ -63,114 +63,29 @@ import Foundation
...
+
+	/// Creates a client which will attempt to send messages to the specified mach service.
+	///
+	/// - Parameters:
+	///   - machServiceName: The name of the XPC mach service; no validation is performed on this.
+	public init(machServiceName: String) {
+		super.init(serviceName: machServiceName)
+	}
+
+	/// Creates and returns connection for the mach service stored by this instance of the client.
```
amomchilov, line 80:
```suggestion
	/// Creates and returns connection for the service stored by this client.
```

Sources/SecureXPC/XPCServiceClient.swift
```diff
@@ -0,0 +1,22 @@
...
+//  XPCServiceClient.swift
+//  
+//
+//  Created by Alexander Momchilov on 2021-11-08.
+//
+
+import Foundation
+
+public class XPCServiceClient: XPCBaseClient {
+	/// Creates and returns connection for the mach service stored by this instance of the client.
```
amomchilov, line 11:
```suggestion
	/// Creates and returns a connection for the XPC service stored by this client.
```

Sources/SecureXPC/XPCServiceServer.swift
```diff
@@ -0,0 +1,29 @@
+//
+//  File.swift
+//  
+//
+//  Created by Alexander Momchilov on 2021-11-07.
+//
+
+import XPC
+
+public class XPCServiceServer: XPCBaseServer {
```
amomchilov, line 10: TODO: Add docs

Sources/SecureXPC/XPCBaseClient.swift
```diff
@@ -0,0 +1,113 @@
+//
+//  XPCBaseClient.swift
+//  SecureXPC
+//
+//  Created by Josh Kaplan on 2021-10-09
+//
+
+import Foundation
+
+public class XPCBaseClient {
```
amomchilov, line 10: TODO: Add docs

Sources/SecureXPC/XPCServiceServer.swift
```diff
@@ -0,0 +1,29 @@
...
+	public static let service = XPCServiceServer()
+
+	private override init() {}
+
+	public override func start() -> Never {
+		xpc_main { connection in
```
jakaplan, line 16: What happens when this is called not from an XPC Service? Considering it's quite confusing which type of XPC to use, I expect people will screw this up. Would be good if we could throw an error when this is used wrong.

jakaplan (commented):
Exciting to see this gain more functionality!

I think this API design is absolutely workable, but because the distinction between an XPC Service and a "service" which vends a Mach service is frankly confusing I think the following might be result in more developers doing the right thing on their first try. It'd also give us more flexibility to change the API in the future without being a breaking change.

- Publicly expose only `XPCClient` and `XPCServer` which lack public initializers
- `XPCServer` has three static functions which all return an `XPCServer`
   - `forXPCService()` which is equivalent to the `XPCServiceService.server` in this PR
   - `forBlessedHelperTool()` as it currently exists in `XPCMachServer`
   - `forMachService(withName: String, clientRequirements: [SecRequirement])` which is equivalent to the current `XPCMachServer(machServiceName: String, clientRequirements: [SecRequirement])` initializer
- `XPCClient` has two static functions which both return `XPCClient`:
   - `forXPCService(withName: String)` which is equivalent to the initializer for the new `XPCServiceClient` in this PR
   - `forMachService(withName: String)` which is equivalent to the existing initializer for `XPCMachClient`

In practice they could return subclasses like the ones that exist in this diff, but that's an implementation detail and we could change that any time in the future without resulting in an incompatible change. This flexibility is also a reason I like this approach.

Thoughts?

Sources/SecureXPC/XPCBaseClient.swift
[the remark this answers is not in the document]
amomchilov: It has to be at least as visible as it's subclasses. If this is internal or private, the subclasses can't be public.


Itll change with your new proposed model

Sources/SecureXPC/XPCMachServer.swift
```diff
@@ -1,8 +1,8 @@
 //
 //  XPCMachServer.swift
-//  SecureXPC
+//  
 //
-//  Created by Josh Kaplan on 2021-10-09
+//  Created by Alexander Momchilov on 2021-11-08.
```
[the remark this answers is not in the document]
amomchilov: I renamed the subclass to be the superclass, then made a new file for the subclass. See the git commits  for history

Sources/SecureXPC/XPCServiceServer.swift
amomchilov: Good question. I couldn't find any way to check if a program is executing as an XPC service. Do you know of any way to do that?

amomchilov:
> Exciting to see this gain more functionality!
> 
> I think this API design is absolutely workable, but because the distinction between an XPC Service and a "service" which vends a Mach service is frankly confusing I think the following might result in more developers doing the right thing on their first try. It'd also give us more flexibility to change the API in the future without it being a breaking change.
> 
>     * Publicly expose only `XPCClient` and `XPCServer` which lack public initializers
> 
>     * `XPCServer` has three static functions which all return an `XPCServer`
>       
>       * `forXPCService()` which is equivalent to the `XPCServiceService.server` in this PR
>       * `forBlessedHelperTool()` as it currently exists in `XPCMachServer`
>       * `forMachService(withName: String, clientRequirements: [SecRequirement])` which is equivalent to the current `XPCMachServer(machServiceName: String, clientRequirements: [SecRequirement])` initializer
> 
>     * `XPCClient` has two static functions which both return `XPCClient`:
>       
>       * `forXPCService(withName: String)` which is equivalent to the initializer for the new `XPCServiceClient` in this PR
>       * `forMachService(withName: String)` which is equivalent to the existing initializer for `XPCMachClient`
> 
> 
> In practice they could return subclasses like the ones that exist in this diff, but that's an implementation detail and we could change that any time in the future without resulting in an incompatible change. This flexibility is also a reason I like this approach.
> 
> Thoughts?

I love it! I'll play around with it in the coming days.

Only nit: I think `with` is mildly discouraged in parameter labels. I think ``forXPCService(named:)` reads nicer.

Sources/SecureXPC/XPCBaseClient.swift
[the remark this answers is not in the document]
jakaplan: Oh interesting I didn't realize that was a Swift constraint. I've rarely created class inheritance like this in Swift so I just don't think I've ever encountered it. (Java for contrast allows package private — roughly equivalent to `internal` — superclasses with `public` subclasses.) Being able to hide this abstraction makes me like the model I proposed even more.

Sources/SecureXPC/XPCServiceServer.swift
jakaplan: Searching around a bit I couldn't find a publicly documented way to determine this; however, I think what we want is actually the opposite and that is partially possible enough to be helpful.

It _should be_ possible to determine in the running code is *not* an XPC Service. Namely we can check if the `CFBundlePackageType` Info.plist entry is present and equal to "XPC!". If that's not the case, it's not an XPC Service. (We could go one step further and find the bundle location and see if it's where it should be within the containing app bundle, but not sure that's worth the extra complexity.) When it's not an XPC Service, throwing an error with a useful debug message would be quite helpful.

jakaplan:
> I love it! I'll play around with it in the coming days.

Awesome. I haven't thought it all through, but I'm also hopeful this architecture will extend nicely for passing connections between processes.

> Only nit: I think `with` is mildly discouraged in parameter labels. I think ``forXPCService(named:)` reads nicer.

Good to know. `named` does read well.

amomchilov:
> will extend nicely for passing connections between processes.

Spoilers! 🤫

(I'm working on this next)

I think the class hierarchy actually might need to be re-jigged to "secure" vs "insecure" (client/server), instead of "mach service" vs "xpc service".

This is because from what I can tell, there's no way to know what kind of service is modelled by a `xpc_endpoint_t`. Even once you hydrate it back into an `xpc_connection_t`, I still don't know of a way to distinguish which kind of service it's connected to.

Sources/SecureXPC/XPCServiceServer.swift
amomchilov: Great find!

jakaplan:
> I think the class hierarchy actually might need to be re-jigged to "secure" vs "insecure" (client/server), instead of "mach service" vs "xpc service".
> 
> This is because from what I can tell, there's no way to know what kind of service is modelled by a `xpc_endpoint_t`. Even once you hydrate it back into an `xpc_connection_t`, I still don't know of a way to distinguish which kind of service it's connected to.

This is only relevant for the client right? Since the client doesn't play a role in the security, does this distinction need to exist? I could envision under the hood there's three subclasses of `XPCClient`:
- Mach: creates a new mach connection 
- XPC Service: creates a new "regular" connection
- Transferred (not sure the best name for this): creates connection by hydrating `xpc_endpoint_t`

And I'm imagining there wouldn't be a way for a user to directly. create a "transferred" client but it's instead somehow returned from a call to the server (hand wave on specifically how that happens).

Sources/SecureXPC/XPCServiceServer.swift
jakaplan: Yeah I have to imagine the XP Service doesn't work, but I haven't personally tried it. The [documentation](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingXPCServices.html#//apple_ref/doc/uid/10000172i-SW6-SW6) says this key/value pair is required:
> XPC requires you to specify a number of special key-value pairs in the Info.plist file within the service helper’s bundle. These keys are listed below.

amomchilov: I'm opting to use [`CFBundleGetPackageInfo`](https://developer.apple.com/documentation/corefoundation/1537089-cfbundlegetpackageinfo#) for this, because it's public API (rather than relying on the info.plist directly. I doubt the format will change, but why bother risking it).

Here's what that might look like:

```swift
extension Bundle {
	func getPackageInfo() -> (packageType: String?, pacakgeCreator: String?) {
		var packageTypeI = UInt32()
		var packageCreatorI = UInt32()

		CFBundleGetPackageInfo(CFBundleGetMainBundle(), &packageTypeI, &packageCreatorI)

		packageTypeI = packageTypeI.bigEndian
		packageCreatorI = packageCreatorI.bigEndian

		func uint32ToString(_ input: UInt32) -> String? {
			if input == 0 { return nil }
			var input = input
			return String(data: Data(bytes: &input, count: 4), encoding: .utf8)
		}

		return (uint32ToString(packageTypeI), uint32ToString(packageCreatorI))
	}

	func isXPCService() -> Bool {
		self.getPackageInfo().packageType == "XPC!"
	}
}
```

In case you were curious, the actual [implementation of it](https://github.com/apple/swift-corelibs-foundation/blob/512e4121415677c9c85b69eac5a2109815243861/CoreFoundation/PlugIn.subproj/CFBundle_InfoPlist.c#L964-L971) is [totally bizarre](https://github.com/apple/swift-corelibs-foundation/blob/512e4121415677c9c85b69eac5a2109815243861/CoreFoundation/PlugIn.subproj/CFBundle_InfoPlist.c#L850-L958).

amomchilov:
> Since the client doesn't play a role in the security, does this distinction need to exist?

I'm not sure. In principle, XPC allows an "inverse" connection to be opened, where the *service* can initiate message sends to the app (rather than just replying). Is that something we'd ever want? (I have no need for it in my app, that I can envision)

> There needs to be 3 client classes

Yeah, that makes perfect sense. They each override `createConnection` to use the appropriate `xpc_connection_create` API.

> And I'm imagining there wouldn't be a way for a user to directly. create a "transferred" client but it's instead somehow returned from a call to the server (hand wave on specifically how that happens).

Meh, if they have an `xpc_endpoint_t`, I don't see why we shouldn't allow them be able to create our client from it.

Sources/SecureXPC/XPCBaseClient.swift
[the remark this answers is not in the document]
amomchilov: > Java for contrast allows package private — roughly equivalent to internal — superclasses with public subclasses.

Huh, I used Java for a while and I didn't even know that.

> Being able to hide this abstraction makes me like the model I proposed even more.

Agreed.

Sources/SecureXPC/XPCBaseClient.swift
```diff
@@ -0,0 +1,113 @@
...
+    
+    // This client implementation intentionally does not store a reference to the xpc_connection_t as it can become
+    // invalid for numerous reasons. Since it's expected relatively few messages will be sent and the lowest possible
+    // latency isn't needed, it's simpler to always create the connection on demand each time a message is to be sent.
+    
+    internal let serviceName: String
```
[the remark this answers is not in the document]
amomchilov: What makes you say that? I think `serviceName` is pretty spot on. Either it's the XPC service's name, or the Mach service's name.

The connection, strictly speaking, doesn't have a name. The service it connects to does.

Sources/SecureXPC/XPCMachClient.swift
```diff
@@ -63,114 +63,29 @@ import Foundation
...
-            // A block *must* be set as the handler, even though this block does nothing.
-            // If it were not set, a crash would occur upon calling xpc_connection_resume.
-        })
-        xpc_connection_resume(connection)
-        
-        return connection
-    }
+public class XPCMachClient: XPCBaseClient {
+
+	// This client implementation intentionally does not store a reference to the xpc_connection_t as it can become
```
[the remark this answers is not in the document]
amomchilov: I'll move it to the `createConnection()` method, where I think it'll be closer to where it'll be expected.

I think keeping it replicated between the two kinds of client is kinda worth it, in case we change this behaviour on a case-by-case basis. 

WDYT?

Sources/SecureXPC/XPCMachServer.swift
```diff
@@ -76,413 +76,199 @@ import Foundation
...
-        
-        return auditToken
-    }
-    
-    private func handleMessage(connection: xpc_connection_t, message: xpc_object_t, reply: inout xpc_object_t?) throws {
-        let request = try Request(dictionary: message)
+		self.machService = machServiceName.withCString { serviceNamePointer in
+			return xpc_connection_create_mach_service(serviceNamePointer,
+													  nil,
+													  UInt64(XPC_CONNECTION_MACH_SERVICE_LISTENER))
```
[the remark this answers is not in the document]
amomchilov: Xcode has a mind of its own. I'll fix it up.

amomchilov:
@jakaplan I'm pretty with how this is shaping up. I think it's ready for review and merge.

I updated the documentation to fix the broken references, but I didn't document this XPC service capability much. Could I leave that for you to put in? I don't know how you would like to organize that information. (Should it all reside on `XPCClient` and `XPCService`?)

Sources/SecureXPC/XPCBaseClient.swift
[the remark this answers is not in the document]
jakaplan: Fair points, I take that back.

Sources/SecureXPC/XPCServiceServer.swift
jakaplan: Using a direct API to access the information is sensible; also reduce the amount of "magic" constants used. This looks great!

I'm not that surprised the implementation for this is so gnarly, that's what a couple decades worth of backwards compatibility and workarounds often end up looking like 😬

If this really only makes sense to be used by the `XPCServiceServer`, seems preferable to not have it be an extension of `Bundle`. This is a reflection of my own architectural bias which is to always default to limited access/scope and only increase it once the need arises; but I don't insist that's the case so long as this remains framework internal.

When this actually make it into the code, would be good to have a comment referencing where the "XPC!" string comes from; namely: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingXPCServices.html#//apple_ref/doc/uid/10000172i-SW6-SW6

Nit: `pacakgeCreator` -> `packageCreator`

jakaplan:
> In principle, XPC allows an "inverse" connection to be opened, where the _service_ can initiate message sends to the app (rather than just replying). Is that something we'd ever want? (I have no need for it in my app, that I can envision)

Not sure if it should be part of the v1 as it certainly adds quite a bit of complexity. It's on my radar as I'm going to need a Launch Agent to regularly update an app with information. However, to start with I might just approach this with polling and that approach very well may turn out to be good enough for my purposes. What I'm working on has no high performance/low latency needs at all.

> Meh, if they have an `xpc_endpoint_t`, I don't see why we shouldn't allow them be able to create our client from it.

I feel rather strongly this not be the case. I don't want SecureXPC to feel like a wrapper of the C API. I want it to feel like the way to use XPC with Swift (well I originally envisioned it as the way to use Swift with XPC Mach services, while this PR has made a great expansion on that). As a point of reference, the Objective-C XPC API does not offer any way to access the C API even though I very much expect it uses it under the hood (although I could be wrong).

Sources/SecureXPC/Client/XPCMachClient.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+
+import Foundation
+
+internal class XPCMachClient: XPCClient {
+	/// Creates a client which will attempt to send messages to the specified mach service.
+	///
+	/// - Parameters:
+	///   - machServiceName: The name of the XPC mach service; no validation is performed on this.
+	public init(machServiceName: String) {
```
jakaplan, line 15: Is there a reason to have this initializer? It should inherit `XPCClient`'s and since all this is doing is calling that one, don't believe this serves a purpose.

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -63,152 +67,91 @@ import Foundation
...
+	/// Because many processes on the system can talk to an XPC Mach Service, when creating a server it is required that you specifiy the security requirements
+	/// of any connecting clients:
+	/// ```swift
+	/// let reqString = """identifier "com.example.AuthorizedClient" and certificate leaf[subject.OU] = "4L0ZG128MM" """
+	/// var requirement: SecRequirement?
+	/// if SecRequirementCreateWithString(reqString as CFString,
+	///                                   SecCSFlags(),
+	///                                   &requirement) == errSecSuccess,
+	///   let requirement = requirement {
+	///    let server = XPCMachServer(machServiceName: "com.example.service",
```
jakaplan, line 99: This is now out of date. I can address this when updating documentation after this has been merged.

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -63,152 +67,91 @@ import Foundation
...
+	/// Incoming requests will be accepted from clients that meet _any_ of the `SMAuthorizedClients` requirements.
+	///
+	/// > Important: No requests will be processed until ``start()`` is called.
+	///
+	/// - Returns: A server instance initialized with the embedded property list entries.
+	public static func forBlessedHelperTool() throws -> XPCServer {
+		try XPCMachServer._forBlessedHelperTool()
+	}
+
+	public static func forThisXPCService() -> XPCServer {
```
jakaplan, line 140: Why the `This` in the function name?

I'd like us to be consistent in the naming for this and the above function so either:
 - `forThisBlessedHelperTool()` and `forThisXPCService()`; OR
 - `forBlessedHelperTool()` and `forXPCService()`

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -261,25 +204,7 @@ public class XPCMachServer {
...
-            })
-            xpc_connection_resume(connection)
-        })
-        xpc_connection_resume(self.machService)
-        
-        dispatchMain()
-    }
-    
-    private func handleEvent(connection: xpc_connection_t, event: xpc_object_t) {
+    internal func handleEvent(connection: xpc_connection_t, event: xpc_object_t) {
```
jakaplan, line 207: Is there a reason this shouldn't remain `private`?

Sources/SecureXPC/Server/XPCServiceServer.swift
```diff
@@ -0,0 +1,29 @@
...
+//  XPCServiceServer.swift
+//  
+//
+//  Created by Alexander Momchilov on 2021-11-07.
+//
+
+import XPC
+
+internal class XPCServiceServer: XPCServer {
+	public static let service = XPCServiceServer()
```
jakaplan, line 11: It doesn't make any difference since the class is `internal`, but make this `internal` as well for clarity?

Sources/SecureXPC/Server/XPCMachServer.swift
```diff
@@ -0,0 +1,167 @@
...
+
+		self.machService = machServiceName.withCString { serviceNamePointer in
+			return xpc_connection_create_mach_service(
+				serviceNamePointer,
+				nil, // targetq: DispatchQueue, defaults to using DISPATCH_TARGET_QUEUE_DEFAULT
+				UInt64(XPC_CONNECTION_MACH_SERVICE_LISTENER))
+		}
+	}
+
+	internal static func _forBlessedHelperTool() throws -> XPCMachServer {
```
jakaplan, line 25: I understand why the the leading `_` is here, but it seems rather un-Swift like? Maybe call it `createForBlessedHelperTool()`? I'm not horribly opposed to the underscore, it just looks odd to me.

jakaplan (changes requested):
This is look really close! The only major thing remaining is getting consistent naming for the `XPCServer` static factory functions.

jakaplan:
> I updated the documentation to fix the broken references, but I didn't document this XPC service capability much. Could I leave that for you to put in? I don't know how you would like to organize that information. (Should it all reside on `XPCClient` and `XPCService`?)

Yup, happy to do so.

Sources/SecureXPC/XPCServiceServer.swift
amomchilov: > seems preferable to not have it be an extension of `Bundle`

Ah, I thought it seemed quite fitting. It's what I would do in my own project, but IDK if the rules change once you're writing a library that others integrate, because:

> so long as this remains framework internal.

I've been putting off testing/researching this: Is there UB if two modules define the same internal extensions on the same type? I'll look into it.

Sources/SecureXPC/Client/XPCMachClient.swift
amomchilov: Nope! ✂️

Sources/SecureXPC/Server/XPCServer.swift
amomchilov: I think I like `forThisBlessedHelperTool()`. WDYT?

Sources/SecureXPC/Server/XPCServer.swift
amomchilov: It's called from both subclass implementations of `start()`.

Sources/SecureXPC/Server/XPCMachServer.swift
amomchilov: I needed some arbitrary distinction to make it not clash with the inherited `forBlessedHelperTool()`.

Alternatively, I could make the parent as `class` (instead of `static`, which is like `final class`), but I didn't want that to be exposed to public API

amomchilov:
Hey Jake, I've incorporated your feedback.

Here's the outstanding stuff before merge. Feel free to edit this PR to your personal taste:

1. [ ] Decide between ([comment](https://github.com/trilemma-dev/SecureXPC/pull/10#discussion_r746391022)):
    * `forThisBlessedHelperTool()` and `forThisXPCService()`; OR
    * `forBlessedHelperTool()` and `forXPCService()`
2. [ ] Solve [the naming problem around `_forBlessedHelperTool()`](https://github.com/trilemma-dev/SecureXPC/pull/10#discussion_r746393774)

Post merge:

1. [ ] [Update the docs](https://github.com/trilemma-dev/SecureXPC/pull/10#discussion_r746391022)

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Upon re-reading this portion of the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) :
> Include all the words needed to avoid ambiguity for a person reading code where the name is used.

I concur, because the inclusion of the "This" reduces ambiguity.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Ah whoops, not sure I how I forgot that.

jakaplan (approved):

jakaplan:
> Hey Jake, I've incorporated your feedback.

FYI my name is Josh. The "a" is for my middle name because I "jkaplan" was already taken :/

> Here's the outstanding stuff before merge. Feel free to edit this PR to your personal taste:

Went with what you had, didn't modify the PR.

> Post merge:
> 
> 1. [ ]  [Update the docs](https://github.com/trilemma-dev/SecureXPC/pull/10#discussion_r746391022)

Will start working on this. May hold off on updating the README.md until a new release is tagged.

PR #18, opened 2021-11-13
Add xpc client.error handler

### LINKED ISSUE #17: Add support for handling client errors on reply-less messages
When there's no "reply" block, there's nowhere to detect errors that were encountered after sending a reply-less message.

amomchilov:
Fixes #17

### DIFF
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -126,9 +126,10 @@ public class XPCClient {
         XPCMachClient(serviceName: machServiceName)
     }
 
-	// MARK: Implementation
+    // MARK: Implementation
 
     internal let serviceName: String
+    private var connection: xpc_connection_t? = nil
     
     /// Creates a client which will attempt to send messages to the specified mach service.
     ///
@@ -222,15 +223,25 @@ public class XPCClient {
     }
 
     private func getConnection() -> xpc_connection_t {
-        let newConnection = self.createConnection()
+        if let existingConnection = self.connection { return existingConnection }
 
-        xpc_connection_set_event_handler(newConnection, { (event: xpc_object_t) in
-            // A block *must* be set as the handler, even though this block does nothing.
-            // If it were not set, a crash would occur upon calling xpc_connection_resume.
-        })
+        let newConnection = self.createConnection()
+        self.connection = newConnection
+
+        xpc_connection_set_event_handler(newConnection, self.handle(connectionEvent:))
         xpc_connection_resume(newConnection)
 
         return newConnection
+    }
+
+    private func handle(connectionEvent: xpc_object_t) {
+        if xpc_equal(connectionEvent, XPC_ERROR_CONNECTION_INVALID) {
+            self.connection = nil
+        } else if xpc_equal(connectionEvent, XPC_ERROR_CONNECTION_INTERRUPTED) {
+            self.connection = nil
+        } else if xpc_equal(connectionEvent, XPC_ERROR_TERMINATION_IMMINENT) {
+            self.connection = nil
+        }
     }
 
     // MARK: Abstract methods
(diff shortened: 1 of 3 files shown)


Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -221,9 +223,35 @@ public class XPCClient {
...
+
+        xpc_connection_set_event_handler(newConnection, self.handle(event:))
+        xpc_connection_resume(newConnection)
+
+        return newConnection
+    }
+
+    private func handle(event: xpc_object_t) {
+		// This callback is *not* called for "reply" messages,
+		// but *is* called when events happens when reply-less messages are sent
```
jakaplan, line 237: So that means if an API user sets `errorHandler` then it's only called for reply-less messages? That seems like a deal breaker due to its inconsistent behavior. While the current behavior means issue #17 exists, it's very clear what the behavior is since there's no way to get back an error for the `send(...)` functions.

Seems like a way to ensure consistency here is under the hood to always send XPC messages that expect a reply and then have the `send(...)` functions that don't have a reply be able to provide an error handler. Or potentially as a bonus make it a completion handler so that it's possible for the client to get confirmation the message was received.

Aside from the consistency issues, the approach in this PR makes it impossible (or at least inconvenient?) for the client to know _which_ `send(...)` calls didn't succeed.

jakaplan (changes requested):
The simplification for `createConnection()` is nice!

I understand the desire to address issue #17, but if my understanding is correct based on the code comments then this approach is too inconsistent compared to current behavior.

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -221,9 +223,35 @@ public class XPCClient {
         })
     }
 
-	// MARK: Abstract methods
+    private func getConnection() -> xpc_connection_t {
```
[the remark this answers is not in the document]
amomchilov: Hah, this is a spoiler for a future where I persist the connection. The latency of starting one up every time is far too high for my application.

https://github.com/amomchilov/SecureXPC/commit/b0f2bf4978cde07d953302136ce61c994e1a7312

Sources/SecureXPC/Client/XPCClient.swift
amomchilov: Hmmm you're totally right.

> Seems like a way to ensure consistency here is under the hood to always send XPC messages that expect a reply and then have the send(...) functions that don't have a reply be able to provide an error handler. Or potentially as a bonus make it a completion handler so that it's possible for the client to get confirmation the message was received.

I experimented with this idea. One thing to note is that you can't just use a `Result<(), XPCError>` reply type as usual (because `Void` isn't `Codable`).

> Or potentially as a bonus make it a completion handler so that it's possible for the client to get confirmation the message was received.

I think this would be good. The completion handler can be `((XPCError?) -> Void)?`. When it's `nil`, we can optimize and use the reply-less API.

> Aside from the consistency issues, the approach in this PR makes it impossible (or at least inconvenient?) for the client to know which send(...) calls didn't succeed.

True. I didn't notice this because I was narrowly interested in seeing the connected/disconnected status of my connection (I built a small little debugging UI in my App using [`NSImageNameStatusNone`/`NSImageNameStatusAvailable`/`NSImageNameStatusUnavailable`](https://hetima.github.io/fucking_nsimage_syntax/) indicators)

amomchilov: Actually, could we move this design-related discussion thread to the main issue? #17

jakaplan:
Has anything changed here since the prior commit? Github seems to think not and doing a manual review also seems to indicate that (but my memory is certainly fallible), so wanted to explicitly ask in case I'm overlooking something.

amomchilov:
Ah, I forgot I had a PR open for this. I was just rebasing all my branches.

There's no change here, I'm still exploring this problem area, though.

jakaplan:
@amomchilov I think this PR can be abandoned?

PR #21, opened 2021-11-28
Persistent client connections

amomchilov:
I think it's time to add the ability to persist client connections. This unlocks several capabilities:

1. Passing a connection between processes (`xpc_endpoint_t`) requires that you create an endpoint from a connection, and *not* merely a service name.
2. This hugely reduces the latency of subsequent messages sent to the same connection (which is crucial for my use case, where I'm sending several messages a second, and the connection setup time is well over a second)
3. Adding support for `xpc_endpoint_t` (coming soon™) will mean that anonymous connections will be possible. I can use this to [write an integration test](https://developer.apple.com/forums/thread/693955) that does a full test of the client messaging/sending code

This code might seem familiar, and that's because this commit used to be a part of #18. It can ship on its own here without any of the other error-handling (I'm still working on that)

### DIFF
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -148,7 +148,7 @@ public class XPCClient {
     /// - Throws: If unable to encode the route. No error will be thrown if communication with the server fails.
     public func send(route: XPCRouteWithoutMessageWithoutReply) throws {
         let encoded = try Request(route: route.route).dictionary
-        xpc_connection_send_message(createConnection(), encoded)
+        xpc_connection_send_message(getConnection(), encoded)
     }
     
     /// Sends a message which will not receive a response.
@@ -159,7 +159,7 @@ public class XPCClient {
     /// - Throws: If unable to encode the message or route. No error will be thrown if communication with the server fails.
     public func sendMessage<M: Encodable>(_ message: M, route: XPCRouteWithMessageWithoutReply<M>) throws {
         let encoded = try Request(route: route.route, payload: message).dictionary
-        xpc_connection_send_message(createConnection(), encoded)
+        xpc_connection_send_message(getConnection(), encoded)
     }
     
     /// Sends with no message and provides the reply as either a message on success or an error on failure.
@@ -191,7 +191,7 @@ public class XPCClient {
     /// Does the actual work of sending an XPC message which receives a reply.
     private func sendWithReply<R: Decodable>(encoded: xpc_object_t,
                                              withReply reply: @escaping XPCReplyHandler<R>) {
-        xpc_connection_send_message_with_reply(createConnection(), encoded, nil, { xpcResponse in
+        xpc_connection_send_message_with_reply(getConnection(), encoded, nil, { xpcResponse in
             let result: Result<R, XPCError>
             if xpc_get_type(xpcResponse) == XPC_TYPE_DICTIONARY {
                 do {
@@ -221,9 +221,21 @@ public class XPCClient {
         })
     }
 
-	// MARK: Abstract methods
+    private func getConnection() -> xpc_connection_t {
+        let newConnection = self.createConnection()
 
-	/// Creates and returns a connection for the service represented by this client.
+        xpc_connection_set_event_handler(newConnection, { (event: xpc_object_t) in
+            // A block *must* be set as the handler, even though this block does nothing.
+            // If it were not set, a crash would occur upon calling xpc_connection_resume.
+        })
+        xpc_connection_resume(newConnection)
+
+        return newConnection
+    }
+
+    // MARK: Abstract methods
+
+    /// Creates and returns a connection for the service represented by this client.
     internal func createConnection() -> xpc_connection_t {
         fatalError("Abstract Method")
     }
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -126,9 +126,10 @@ public class XPCClient {
         XPCMachClient(serviceName: machServiceName)
     }
 
-	// MARK: Implementation
+    // MARK: Implementation
 
     internal let serviceName: String
+    private var connection: xpc_connection_t? = nil
     
     /// Creates a client which will attempt to send messages to the specified mach service.
     ///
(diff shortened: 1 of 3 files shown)


Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -126,9 +126,10 @@ public class XPCClient {
         XPCMachClient(serviceName: machServiceName)
     }
 
-	// MARK: Implementation
+    // MARK: Implementation
 
     internal let serviceName: String
```
jakaplan, line 131: Should this variable still exist in the super class? While both subclasses currently do make use of it, seems a bit off from an encapsulation point of view as this class itself does not. (I realize this isn't new behavior as of this PR, but just noticed it now.)

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -221,9 +222,24 @@ public class XPCClient {
         })
     }
 
-	// MARK: Abstract methods
+    private func getConnection() -> xpc_connection_t {
```
jakaplan, line 225: If I'm understanding this correctly, once the connection gets invalidated (`XPC_ERROR_CONNECTION_INVALID`) the client becomes unusable. We shouldn't make users of this API have to manually handle this when we can automatically do so under the hood.

This is one of the major reasons the current design doesn't persist the connection (although as you note that of course that comes with a significant performance penalty) - by not doing so it always gracefully recovers from this situation. With the current design of this PR, I think it'd make sense to properly receive all of the XPC errors and then `nil` out the connection as needed.

jakaplan (changes requested):

Sources/SecureXPC/Client/XPCClient.swift
amomchilov: Sure, could you open a new issue about refactoring this?

Sources/SecureXPC/Client/XPCClient.swift
amomchilov: Ah yes, you're right. I need to pick off some parts of https://github.com/trilemma-dev/SecureXPC/pull/18/files to do this (introducing an error handler for the sake of the client's connection, not for the errors encounters on specific messages)

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: Actually looking at #23 I think this should be addressed, see that PR for relevant comments.

amomchilov:
Is there anything else that would need to be done besides this? https://github.com/trilemma-dev/SecureXPC/compare/53f1526a9b264b7d68ed8f37c7002c66b53daeb1..47ba03aeb59e5a3134becb80554743cad89827ba

jakaplan:
> Is there anything else that would need to be done besides this? https://github.com/trilemma-dev/SecureXPC/compare/53f1526a9b264b7d68ed8f37c7002c66b53daeb1..47ba03aeb59e5a3134becb80554743cad89827ba
> 
> I don't have a really comprehensive way of testing this, but it seems to work in my app.

I believe it's also necessary to `nil` out the connection in response to an unsuccessful message send with reply:
https://github.com/trilemma-dev/SecureXPC/blob/47ba03aeb59e5a3134becb80554743cad89827ba/Sources/SecureXPC/Client/XPCClient.swift#L212-L218

jakaplan (changes requested):
Once the connection has been set to `nil` on unsuccessful send with reply, then this should be good to go. I wish I knew of a more comprehensive way to test this.

amomchilov:
Are you sure that those branches ever even get hit? IIRC, those connection events never hit the reply block, only the connection'event handler.

I can confirm later today

jakaplan:
> Are you sure that those branches ever even get hit? IIRC, those connection events never hit the reply block, only the connection'event handler.


From the documentation for [`xpc_connection_send_message_with_reply`](https://developer.apple.com/documentation/xpc/1448795-xpc_connection_send_message_with):
> `handler` The handler block to invoke when a reply to the message is received from the connection. If the remote service exits prematurely before the reply was received, the `XPC_ERROR_CONNECTION_INTERRUPTED` error will be returned. If the connection went invalid before the message could be sent, the `XPC_ERROR_CONNECTION_INVALID` error will be returned.

So handling the `XPC_ERROR_TERMINATION_IMMINENT` is unnecessary (which makes sense as this only applies to servers), but otherwise - yes the other two cases can be hit.

amomchilov:
> If the remote service exits prematurely before the reply was received, the `XPC_ERROR_CONNECTION_INTERRUPTED` error will be returned.

Hmmm, I can't recreate that. I have a route that looks like this:

```
let quitRoute = XPCRouteWithoutMessageWithoutReply("quit")
service.registerRoute(quitRoute, handler: {
	exit(EXIT_SUCCESS)
})
```

It doesn't cause `XPC_ERROR_CONNECTION_INTERRUPTED` on the reply :/

amomchilov:
I updated to always refresh the connection on any error.

This isn't exactly ideal because:

1. I don't really understand these errors, and handling errors you don't understand is like ... bad 😅 
2. IIRC, an invalidated connection can't be reconnected to, but as-implemented, this code will try to anyway. This is on par with the existing behaviour (connect every time no matter what). Perhaps we should revisit in a new issue.

I think this should be good to go now.

jakaplan:
> > If the remote service exits prematurely before the reply was received, the `XPC_ERROR_CONNECTION_INTERRUPTED` error will be returned.
> 
> Hmmm, I can't recreate that. I have a route that looks like this:
> 
> ```
> let quitRoute = XPCRouteWithoutMessageWithoutReply("quit")
> service.registerRoute(quitRoute, handler: {
> 	exit(EXIT_SUCCESS)
> })
> ```
> 
> It doesn't cause `XPC_ERROR_CONNECTION_INTERRUPTED` on the reply :/

That makes sense because there's no reply?

jakaplan:
> 2\. IIRC, an invalidated connection can't be reconnected to, but as-implemented, this code will try to anyway. This is on par with the existing behaviour (connect every time no matter what). Perhaps we should revisit in a new issue.

The connection can no longer be used, but a _new_ connection can be established. Which is what this code does so I don't think there's a logic issue here.

jakaplan (approved):

amomchilov:
> That makes sense because there's no reply?

Kind of, but then I'm not sure what exactly it means for the connection to be interrupted.

Do you think `XPC_ERROR_CONNECTION_INTERRUPTED` would be raised if the process exited *after* it made a reply, but before sending it?

jakaplan:
> > That makes sense because there's no reply?
> 
> Kind of, but then I'm not sure what exactly it means for the connection to be interrupted.

Your code is:
```swift
let quitRoute = XPCRouteWithoutMessageWithoutReply("quit")
service.registerRoute(quitRoute, handler: {
	exit(EXIT_SUCCESS)
})
```

Which means on the client side I expect you're calling:
```swift
public func send(route: XPCRouteWithoutMessageWithoutReply) throws {
    let encoded = try Request(route: route.route).dictionary
    xpc_connection_send_message(createConnection(), encoded)
}
```

That's using the `xpc_connection_send_message` function which does not have a handler. (This is contrast to the `xpc_connection_send_message_with_reply` function which does.) So all of this matches expected behavior unless I'm misunderstanding something.

amomchilov:
Haha oh, duh! By "there's no reply", I thought you meant because my closure was never returning, no reply woudl be sent.

I see what you mean, there's no reply, *at all*.

Issue #22, opened 2021-11-28
Add real XPC integration tests


amomchilov:
I would like to contribute an integration test suite, which exercises the real client/server messaging functions. I would like to build this prior to making pretty big changes as part of #17 and #20.

As it turns out, this can be done without the need to spin up XPC or Mach services, using an anonymous XPC connection, whose ends both run within the same process. Quinn "The Eskimo!" documents how to do this with the NSXPCConnection APIs in [Testing and Debugging XPC Code With an Anonymous Listener](https://developer.apple.com/forums/thread/693955).

I found some more information about anonymous XPC connections in the C XPC APIs in [WWDC 2013 - Session 702 - Efficient design with XPC](http://devstreaming-cdn.apple.com/videos/wwdc/2013/702xfx2xmlrics5pyrjfwu2m/702/702-SD.mov?dl=1) ([slides](devstreaming-cdn.apple.com/videos/wwdc/2013/702xfx2xmlrics5pyrjfwu2m/702/702.pdf?dl=1)), at around 18:33 to 21:50, slides 47-56.

I've split the work into multiple PRs, each one building upon the previous

- [ ] #21
- [ ] #23
- [ ] #24
- [ ] #25

amomchilov:
@jakaplan Hey Josh.

As you can see, this work involves quite a few different parts. It looks like GitHub doesn't have a very good way to handle dependant work like this, so this is a bit rough.

I've split them into their own PRs. For each one, please:

1) Review in the order they're listed above
2) Put comments on the PR that *first* introduces the effected line(s).
3) Don't ship the PRs until their dependency has merged, and I've re-based them

jakaplan:
I'll do my best to start reviewing these tomorrow night (my time), it'll likely take me a few days to get through them. Things right now are quite busy both personally and professionally.

amomchilov:
No worries! I'm just working off a local copy, so you're not blocking anything.

> Things right now are quite busy both personally and professionally.

Good luck!

Same situation here, hence my sporadic bursts of contributions followed by silence haha.

jeff-h:
Sorry for going off-topic, but I just wanted to thank you both for your work on this. I'm hoping to migrate my XPC code to use SecureXPC and really appreciate the effort going into this. I think a lot of people will benefit from this work over time — it's really needed.

jakaplan:
> Sorry for going off-topic, but I just wanted to thank you both for your work on this.

[...]

Issue #1, opened 2021-10-23
Enable GitHub discussions for the 3 repos


amomchilov:
👋  Hey there!

I have some things I'd like to ask about, that aren't necessarily issues, per-se. Could you enable [GitHub discussions](https://docs.github.com/en/discussions) on these repos so I can post there?

P.S. Thanks for open sourcing these. It caught me just at the right time, when I was about to start refactoring a huge chunk of my apps structs into `@objc` classes, solely to support `NSCoding` for XPC. That would have been a major headache, and I'm really hoping I can leverage this package to stay in swift-struct-land!

PR #23, opened 2021-11-28
Add Endpoint APIs

amomchilov:
This PR builds on top of the code introduced in #21. Do not ship until #21 is merged.

Add a new `XPCEndpoint` wrapper around `xpc_endpoint_t`.

### DIFF
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -126,6 +126,20 @@ public class XPCClient {
         XPCMachClient(machServiceName: machServiceName)
     }
 
+	public static func forEndpoint(_ endpoint: XPCServerEndpoint) -> XPCClient {
+        let connection = xpc_connection_create_from_endpoint(endpoint.endpoint)
+
+        xpc_connection_set_event_handler(connection, { (event: xpc_object_t) in
+            fatalError("It should be impossible for this connection to receive an event.")
+        })
+        xpc_connection_resume(connection)
+
+        switch endpoint.serviceDescriptor {
+        case .xpcService(name: let name): return XPCServiceClient(xpcServiceName: name, connection: connection)
+        case .machService(name: let name): return XPCMachClient(machServiceName: name, connection: connection)
+        }
+    }
+
     // MARK: Implementation
 
     private var connection: xpc_connection_t? = nil
@@ -134,7 +148,8 @@ public class XPCClient {
     ///
     /// - Parameters:
     ///   - serviceName: The name of the XPC service; no validation is performed on this.
-    internal init() {
+    internal init(connection: xpc_connection_t? = nil) {
+        self.connection = connection
     }
     
     /// Receives the result of an XPC send. The result is either an instance of the reply type on success or an ``XPCError`` on failure.
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -106,7 +106,7 @@ public class XPCClient {
     ///   - named: The bundle identifier of the XPC Service.
     /// - Returns: A client configured to communicate with the named service.
     public static func forXPCService(named xpcServiceName: String) -> XPCClient {
-        XPCServiceClient(serviceName: xpcServiceName)
+        XPCServiceClient(xpcServiceName: xpcServiceName)
     }
     
     /// Provides a client to communicate with an XPC Mach service.
@@ -123,20 +123,18 @@ public class XPCClient {
     ///    - named: A key in the `MachServices` entry of the tool's launchd property list.
     /// - Returns: A client configured to communicate with the named service.
     public static func forMachService(named machServiceName: String) -> XPCClient {
-        XPCMachClient(serviceName: machServiceName)
+        XPCMachClient(machServiceName: machServiceName)
     }
 
     // MARK: Implementation
 
-    internal let serviceName: String
     private var connection: xpc_connection_t? = nil
     
     /// Creates a client which will attempt to send messages to the specified mach service.
     ///
     /// - Parameters:
     ///   - serviceName: The name of the XPC service; no validation is performed on this.
-    internal init(serviceName: String) {
-        self.serviceName = serviceName
+    internal init() {
     }
     
     /// Receives the result of an XPC send. The result is either an instance of the reply type on success or an ``XPCError`` on failure.
@@ -248,6 +246,11 @@ public class XPCClient {
     }
 
     // MARK: Abstract methods
+
+
+    public var serviceName: String? {
+        fatalError("Abstract Property")
+    }
 
     /// Creates and returns a connection for the service represented by this client.
     internal func createConnection() -> xpc_connection_t {
--- a/Sources/SecureXPC/Client/XPCMachClient.swift
+++ b/Sources/SecureXPC/Client/XPCMachClient.swift
@@ -11,8 +11,16 @@ import Foundation
 ///
 /// In the case of this framework, the Mach service is expected to be represented by an `XPCMachServer`.
 internal class XPCMachClient: XPCClient {
+    private let machServiceName: String
+
+    override var serviceName: String? { machServiceName }
+
+    internal init(machServiceName: String) {
+        self.machServiceName = machServiceName
+    }
+
     /// Creates and returns a connection for the Mach service represented by this client.
     internal override func createConnection() -> xpc_connection_t {
-        xpc_connection_create_mach_service(self.serviceName, nil, 0)
+        xpc_connection_create_mach_service(self.machServiceName, nil, 0)
     }
 }
--- a/Sources/SecureXPC/Client/XPCMachClient.swift
+++ b/Sources/SecureXPC/Client/XPCMachClient.swift
@@ -15,8 +15,9 @@ internal class XPCMachClient: XPCClient {
 
     override var serviceName: String? { machServiceName }
 
-    internal init(machServiceName: String) {
+    internal init(machServiceName: String, connection: xpc_connection_t? = nil) {
         self.machServiceName = machServiceName
+        super.init(connection: connection)
     }
 
     /// Creates and returns a connection for the Mach service represented by this client.
--- a/Sources/SecureXPC/Server/XPCMachServer.swift
+++ b/Sources/SecureXPC/Server/XPCMachServer.swift
@@ -201,6 +201,12 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
 	}
 
     public override var endpoint: XPCServerEndpoint {
+        fatalError("""
+            The ability to export the endpoint of an XPCServer for a Mach Service is untested, and likely broken.
+
+            See https://github.com/trilemma-dev/SecureXPC/issues/33
+        """)
+
         guard let connection = self.serviceListenerConnection else {
             fatalError("An XPCServer's endpoint can only be retrieved after start() has been called on it.")
         }
--- a/Sources/SecureXPC/Server/XPCMachServer.swift
+++ b/Sources/SecureXPC/Server/XPCMachServer.swift
@@ -14,7 +14,8 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
     
     private let machServiceName: String
     private let clientRequirements: [SecRequirement]
-    
+    private var serviceListenerConnection: xpc_connection_t? = nil
+
     /// This should only ever be called from `getXPCMachServer(...)` so that client requirement invariants are upheld.
     private init(machServiceName: String, clientRequirements: [SecRequirement]) {
         self.machServiceName = machServiceName
@@ -152,16 +153,18 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
                 nil, // targetq: DispatchQueue, defaults to using DISPATCH_TARGET_QUEUE_DEFAULT
                 UInt64(XPC_CONNECTION_MACH_SERVICE_LISTENER))
         }
-        
+
+        self.serviceListenerConnection = machService
+
         // Start listener for the Mach service, all received events should be for incoming connections
-         xpc_connection_set_event_handler(machService, { connection in
-             // Listen for events (messages or errors) coming from this connection
-             xpc_connection_set_event_handler(connection, { event in
-                 self.handleEvent(connection: connection, event: event)
-             })
-             xpc_connection_resume(connection)
-         })
-         xpc_connection_resume(machService)
+        xpc_connection_set_event_handler(machService, { connection in
+            // Listen for events (messages or errors) coming from this connection
+            xpc_connection_set_event_handler(connection, { event in
+                self.handleEvent(connection: connection, event: event)
+            })
+            xpc_connection_resume(connection)
+        })
+        xpc_connection_resume(machService)
     }
     
 	public override func startAndBlock() -> Never {
@@ -196,6 +199,18 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
 
 		return accept
 	}
+
+    public override var endpoint: XPCServerEndpoint {
+        guard let connection = self.serviceListenerConnection else {
+            fatalError("An XPCServer's endpoint can only be retrieved after start() has been called on it.")
+        }
+
+        let endpoint = xpc_endpoint_create(connection)
+        return XPCServerEndpoint(
+            serviceDescriptor: .machService(name: self.machServiceName),
+            endpoint: endpoint
+        )
+    }
 
 	/// Wrapper around the private undocumented function `void xpc_connection_get_audit_token(xpc_connection_t, audit_token_t *)`.
 	///
--- a/Sources/SecureXPC/Server/XPCServer.swift
+++ b/Sources/SecureXPC/Server/XPCServer.swift
@@ -314,6 +314,14 @@ public class XPCServer {
 	internal func acceptMessage(connection: xpc_connection_t, message: xpc_object_t) -> Bool {
 		fatalError("Abstract Method")
 	}
+
+    public var serviceName: String? {
+        fatalError("Abstract Property")
+    }
+
+    public var endpoint: XPCServerEndpoint {
+        fatalError("Abstract Method")
+    }
 }
 
 // MARK: public server protocols
--- a/Sources/SecureXPC/Server/XPCServiceServer.swift
+++ b/Sources/SecureXPC/Server/XPCServiceServer.swift
@@ -11,14 +11,18 @@ import Foundation
 ///
 /// In the case of this framework, the XPC Service is expected to be communicated with by an `XPCServiceClient`.
 internal class XPCServiceServer: XPCServer {
-    
 	private static let service = XPCServiceServer()
+    private var connection: xpc_connection_t? = nil
 
     internal static func _forThisXPCService() throws -> XPCServiceServer {
         // An XPC Service's package type must be equal to "XPC!", see Apple's documentation for details
         // https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingXPCServices.html#//apple_ref/doc/uid/10000172i-SW6-SW6
         if mainBundlePackageInfo().packageType != "XPC!" {
             throw XPCError.notXPCService
+        }
+
+        if Bundle.main.bundleIdentifier == nil {
+            throw XPCError.misconfiguredXPCService
         }
         
         return service
@@ -40,9 +44,16 @@ internal class XPCServiceServer: XPCServer {
 
         return (uint32ToString(packageType.bigEndian), uint32ToString(packageCreator.bigEndian))
     }
-    
+
+    // This is safe to unwrap because it was already checked in ``_forThisXPCService``.
+    private lazy var xpcServiceName: String = Bundle.main.bundleIdentifier!
+
 	public override func startAndBlock() -> Never {
 		xpc_main { connection in
+            // This is an @convention(c) closure, so we can't just capture `self`.
+            // We access the connection via the singleton reference, instead.
+            XPCServiceServer.service.connection = connection
+
 			// Listen for events (messages or errors) coming from this connection
 			xpc_connection_set_event_handler(connection, { event in
 				XPCServiceServer.service.handleEvent(connection: connection, event: event)
@@ -55,4 +66,20 @@ internal class XPCServiceServer: XPCServer {
 		// XPC services are application-scoped, so we're assuming they're inheritently safe
 		true
 	}
+
+    public override var serviceName: String? {
+        xpcServiceName
+    }
+
+    public override var endpoint: XPCServerEndpoint {
+        guard let connection = self.connection else {
+            fatalError("An XPCServer's endpoint can only be retrieved after startAndBlock() has been called on it.")
+        }
+
+        let endpoint = xpc_endpoint_create(connection)
+        return XPCServerEndpoint(
+            serviceDescriptor: .xpcService(name: self.xpcServiceName),
+            endpoint: endpoint
+        )
+    }
 }
--- a/Sources/SecureXPC/XPCServerEndpoint.swift
+++ b/Sources/SecureXPC/XPCServerEndpoint.swift
@@ -1,0 +1,21 @@
+//
+//  XPCServerEndpoint.swift
+//  
+//
+//  Created by Alexander Momchilov on 2021-11-28.
+//
+
+import Foundation
+
+// TODO: make this codable so it can be sent over XPC.
+public struct XPCServerEndpoint {
+    // Technically, an `xpc_endpoint_t` is sufficient to create a new connection, on its own. However, it's useful to
+    // be able to communicate the kind of connection, and its name, so we also store those, separately.
+    internal let serviceDescriptor: XPCServiceDescriptor
+    internal let endpoint: xpc_endpoint_t
+
+    internal init(serviceDescriptor: XPCServiceDescriptor, endpoint: xpc_endpoint_t) {
+        self.serviceDescriptor = serviceDescriptor
+        self.endpoint = endpoint
+    }
+}
--- a/Sources/SecureXPC/XPCServiceDescriptor.swift
+++ b/Sources/SecureXPC/XPCServiceDescriptor.swift
@@ -1,0 +1,13 @@
+//
+//  XPCServiceDescriptor.swift
+//  
+//
+//  Created by Alexander Momchilov on 2021-11-28.
+//
+
+import Foundation
+
+internal enum XPCServiceDescriptor {
+    case xpcService(name: String)
+    case machService(name: String)
+}
--- a/Sources/SecureXPC/Client/XPCServiceClient.swift
+++ b/Sources/SecureXPC/Client/XPCServiceClient.swift
@@ -15,8 +15,9 @@ internal class XPCServiceClient: XPCClient {
 
     override var serviceName: String? { xpcServiceName }
 
-    internal init(xpcServiceName: String) {
+    internal init(xpcServiceName: String, connection: xpc_connection_t? = nil) {
         self.xpcServiceName = xpcServiceName
+        super.init(connection: connection)
     }
     
     /// Creates and returns a connection for the XPC service represented by this client.
--- a/Sources/SecureXPC/Client/XPCServiceClient.swift
+++ b/Sources/SecureXPC/Client/XPCServiceClient.swift
@@ -11,6 +11,14 @@ import Foundation
 ///
 /// In the case of this framework, the XPC Service is expected to be represented by an `XPCServiceServer`.
 internal class XPCServiceClient: XPCClient {
+    private let xpcServiceName: String
+
+    override var serviceName: String? { xpcServiceName }
+
+    internal init(xpcServiceName: String) {
+        self.xpcServiceName = xpcServiceName
+    }
+    
     /// Creates and returns a connection for the XPC service represented by this client.
     internal override func createConnection() -> xpc_connection_t {
         xpc_connection_create(self.serviceName, nil)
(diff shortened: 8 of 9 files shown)


Sources/SecureXPC/XPCServerEndpoint.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+
+import Foundation
+
+// TODO: make this codable so it can be sent over XPC.
+public struct XPCServerEndpoint {
+    // Technically, an `xpc_endpoint_t` is sufficient to create a new connection, on its own. However, it's useful to
+    // be able to communicate the kind of connection, and its name, so we also store those, separately.
+    internal let kind: XPCConnectionType
+    internal let serviceName: String?
```
jakaplan, line 15: In what circumstances is it valid for the name to be `nil`?

jakaplan (changes requested):

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -126,16 +126,36 @@ public class XPCClient {
...
 
-	// MARK: Implementation
+	public static func forEndpoint(_ endpoint: XPCServerEndpoint) -> XPCClient {
+        let connection = xpc_connection_create_from_endpoint(endpoint.endpoint)
 
-    internal let serviceName: String
+        xpc_connection_set_event_handler(connection, { (event: xpc_object_t) in
+            // A block *must* be set as the handler, even though this block does nothing.
+            // If it were not set, a crash would occur upon calling xpc_connection_resume.
+            xpc_connection_cancel(event)
```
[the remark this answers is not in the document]
amomchilov: Good Q. I cargo-culted it from the WWDC sessions I linked to in #22.

I meant to come back to this and figure out why they were doing that. Got any ideas? As far as I can tell, this closure is never called.

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -126,16 +126,36 @@ public class XPCClient {
...
+
+    // MARK: Implementation
+
+    internal let serviceName: String?
+    private var connection: xpc_connection_t? = nil
     
     /// Creates a client which will attempt to send messages to the specified mach service.
     ///
     /// - Parameters:
     ///   - serviceName: The name of the XPC service; no validation is performed on this.
```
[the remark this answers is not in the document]
amomchilov: Ah yes, this doc comment is a mess.

> Additionally both parameters can be nil but if both are is that actually valid?

Yes.

The `serviceName` can be `nil` for anonymous services. For regular (named) XPC services, and Mach services, it needs to be non-nil. Perhaps `serviceName` should be an abstract property, and the subclasses can store their own name variable (which can be `non-nil` for mach services.

When creating a client from an `XPCEndpoint`, the connection is already created, and so the client just uses that. In other cases, `connection` will be `nil`, and the client will create a new connection for itself from the service name.

Sources/SecureXPC/Client/XPCMachClient.swift
```diff
@@ -11,20 +11,8 @@ import Foundation
...
-			// A block *must* be set as the handler, even though this block does nothing.
-			// If it were not set, a crash would occur upon calling xpc_connection_resume.
-		})
-		xpc_connection_resume(connection)
-
-		return connection
-	}
+    /// Creates and returns a connection for the Mach service represented by this client.
+    internal override func createConnection() -> xpc_connection_t {
+        xpc_connection_create_mach_service(self.serviceName!, nil, 0)
```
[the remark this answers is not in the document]
amomchilov: Are you thinking perhaps having out own `machServiceName: String` and then having an `override var serviceName: String? { machServiceName }`?

Sources/SecureXPC/Server/XPCMachServer.swift
```diff
@@ -14,7 +14,8 @@ internal class XPCMachServer: XPCServer {
     
     private let machServiceName: String
     private let clientRequirements: [SecRequirement]
-    
+	private var serviceListenerConection: xpc_connection_t? = nil
```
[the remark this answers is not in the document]
amomchilov: lol, yes I am.

Sources/SecureXPC/Server/XPCMachServer.swift
```diff
@@ -193,6 +196,19 @@ internal class XPCMachServer: XPCServer {
 		return accept
 	}
 
+    public override var endpoint: XPCServerEndpoint {
+        guard let connection = self.serviceListenerConection else {
+            fatalError("You can only create an `endpoint` for an XPCMachServer after starting it with `start()`.")
```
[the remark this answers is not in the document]
amomchilov: What language version are we targeting? Throwing properties are a really new addition (5.4 or 5.5, IDR).

In any case, I don't think throwing an error is particular useful, because catching the error wouldn't be particularly useful. It's a fundamental misuse of the API which needs to be fixed before shipping.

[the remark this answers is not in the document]
amomchilov: > Also do backticks actually get honored in error messages? I didn't realize they did.

Nope, just a habit.

Sources/SecureXPC/Server/XPCServiceServer.swift
```diff
@@ -13,6 +13,7 @@ import Foundation
 internal class XPCServiceServer: XPCServer {
     
 	private static let service = XPCServiceServer()
+    private static var connection: xpc_connection_t? = nil
```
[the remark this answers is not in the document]
amomchilov: That was my first instinct, but I had an issue around the closure to `xpc_main` being `@convention(c)` (so it's not really actually a closure, it can't close over unique state, like `self`), so I couldn't use `self.connection = connection`.

I think this could work if I reference it using `XPCServiceServer.service.connection = connection`. I'll try it out.

Sources/SecureXPC/Server/XPCServiceServer.swift
```diff
@@ -55,4 +58,17 @@ internal class XPCServiceServer: XPCServer {
...
+
+    public override var endpoint: XPCServerEndpoint {
+        guard let connection = Self.connection else {
+            fatalError("You can only create an `endpoint` for an XPCServiceServer after starting it with `start()`.")
+        }
+
+        let endpoint = xpc_endpoint_create(connection)
+        return XPCServerEndpoint(
+            kind: .xpcServiceClient,
+            serviceName: "TODO: implement me",
```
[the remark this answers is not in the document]
amomchilov: > In case you didn't already know, this is the name of the bundle so Bundle.main.bundleIdentifier should suffice

Perfect!

> Although it can in theory be nil so perhaps it makes sense to access and store this during initialization

This would be great. It would allow our `XPCServer`s (of any type) to all have a name. (except for anonymous ones, of course)

Sources/SecureXPC/XPCConnectionType.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  XPCConnectionType.swift
+//  
+//
+//  Created by Alexander Momchilov on 2021-11-28.
+//
+
+import Foundation
+
+public enum XPCConnectionType {
```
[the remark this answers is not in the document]
amomchilov: I think I originally had more methods on it. I'll fold it in.

Sources/SecureXPC/XPCServerEndpoint.swift
amomchilov: iff it's an anonymous service

Sources/SecureXPC/Client/XPCClient.swift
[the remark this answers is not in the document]
jakaplan: I very much don't, it looks rather bizarre  I was hoping you could explain why it was added.

My understanding is `xpc_connection_cancel` expects an `xpc_connection_t` and due to typealiasing its compile time valid to pass in an `xpc_object_t`. However, for a client to receive a connection would be quite unexpected. And then why would it cancel be called on it as soon as its received?

Sources/SecureXPC/Client/XPCClient.swift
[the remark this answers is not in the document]
jakaplan: Actually looking at this in more details, looks like there's no need for `XPCClient` to have a `serviceName` instance variable. This can purely be an internal implementation detail of the `XPCMachClient` and `XPCServiceClient` subclasses and a future `XPCAnonymousClient` (or whatever it's called) subclasses wouldn't have one.

Similarly there doesn't appear to be any benefit to passing the connection into the initializer as it'll get created (or return an existing one for an `XPCEndpoint`) as part of the `createConnection()` function. Is that right?

Sources/SecureXPC/Client/XPCMachClient.swift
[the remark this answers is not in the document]
jakaplan: I don't believe there's any need for the super class to have access to `serviceName`. It appears to never be used. So yes to the first part, but no to the second.

Sources/SecureXPC/XPCServerEndpoint.swift
jakaplan: I think we should structure this as part of the `XPCConnectionType` enum with an associated value such that when a connection must have a connection name, it has one, and when it can't then it's not possible to specify one. That'll let the compiler ensure we've implemented the invariant correctly. Thoughts?

Sources/SecureXPC/Server/XPCMachServer.swift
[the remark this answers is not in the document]
jakaplan: Swift 5.5

But fair point that this is a programming error.

Sources/SecureXPC/Client/XPCMachClient.swift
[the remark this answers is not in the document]
amomchilov: It would be used in debug descriptions of the object, and xpc endpoints (in the future when I add the ability to export a client as an endpoint)

Sources/SecureXPC/XPCServerEndpoint.swift
amomchilov: I’ll play around with it and let you know how it goes!

amomchilov: Oh I remember why I didn’t do this:

1. it complicates the structure of the way this value gets encoded (IIRC Swift added the ability to auth-synthesize the codable conformance for an enumeration with associated values, but I haven’t played with that yet)
2. It would make it impossible to build an XPCEndpoint out of a plain `xpc_endpoint_t`.

The latter comes down to a matter of philosophy. My opinion is that we should allow people to access endpoints, connections, audit tokens, etc, if they want to. I don’t think a framework called “SecureXPC” should be afraid of exposing the fact that is uses XPC under the hood (it’s not a secret ;) ). NSXPCConnection was designed as a more opaque wrapper where everything should “just work”, but as you’ve probably seen, that adds a bit too much smoke and mirrors sometimes.

Sources/SecureXPC/Client/XPCMachClient.swift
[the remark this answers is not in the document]
jakaplan: Is there a reason not to take the same approach as `XPCServer` where each subclass actually implements this and the superclass just has:
```swift
public var endpoint: XPCServerEndpoint {
    fatalError("Abstract Method")
}
```

In regards to debug descriptions, I'd think there'd be advantages to having the subclass do that because each subclass has the most comprehensive set of contextual information.

Sources/SecureXPC/XPCServerEndpoint.swift
jakaplan: In regards to #1, if all of the associated values are `Codable` then conformance is automatically generated by the compiler. [`XPCError`](https://github.com/trilemma-dev/SecureXPC/blob/main/Sources/SecureXPC/XPCError.swift) make use of this; several of its cases have `String` associated values.

In regards to #2, I consider it an intentional design choice of SecureXPC that it does not expose the C API it uses. I don't have any issue with `NSXPCConnection` in regards to its abstractions except that it's an Objective-C API which maps very awkwardly to Swift. My view is escape hatches cause many issues and should only be done as a last resort. Fundamentally I don't view SecureXPC as a wrapper of the XPC C API, I view it as its own XPC API which internally makes use of the XPC C API. (This is quite different than the [Blessed framework](https://github.com/trilemma-dev/Blessed) I open sourced which is intentionally a Swift-friendly wrapper of the Authorization Services and Service Management frameworks.) Happy to discuss this in more depth in the Discussions topic.

Sources/SecureXPC/Client/XPCClient.swift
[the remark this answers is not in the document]
amomchilov: > Actually looking at this in more details, looks like there's no need for `XPCClient` to have a `serviceName` instance variable. This can purely be an internal implementation detail of the `XPCMachClient` and `XPCServiceClient` subclasses and a future `XPCAnonymousClient` (or whatever it's called) subclasses wouldn't have one.

Errrr I disagree. NSXPCConnection and the C APIs both take this approach, where connections are just these magical opaque things with very little exposed API. They happen to print their service name (if known) in their `description`, but they're still annoying to deal with, because e.g. their name isn't visible from the object view in the debugger.

I think having the service name be exposed is easy enough that we should have it. It's just a nice convenience for development. (not that anybody should ever need to like, `switch` over it, or anything like that)

> Similarly there doesn't appear to be any benefit to passing the connection into the initializer as it'll get created (or return an existing one for an XPCEndpoint) as part of the createConnection() function. Is that right?

How will "or return an existing one for an XPCEndpoint" work if the connection isn't passed in?

Sources/SecureXPC/Server/XPCServiceServer.swift
```diff
@@ -55,4 +58,17 @@ internal class XPCServiceServer: XPCServer {
 		// XPC services are application-scoped, so we're assuming they're inheritently safe
 		true
 	}
+
+    public override var endpoint: XPCServerEndpoint {
+        guard let connection = Self.connection else {
+            fatalError("You can only create an `endpoint` for an XPCServiceServer after starting it with `start()`.")
```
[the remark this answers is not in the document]
amomchilov: Dedupe against https://github.com/trilemma-dev/SecureXPC/pull/23#discussion_r760132141

Sources/SecureXPC/Client/XPCClient.swift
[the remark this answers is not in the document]
amomchilov: My usage here is a mistaken representation of what they did in their sample.

The correct equivalent would be `xpc_connection_cancel(connection)`. The idea is:

1. They needed something to hold on to `connection` to prevent arc from cleaning it up. In this case, they made their handler capture it, as a handy way to keep the connection alive until its cancelled by some other means.
2. They never expected this block to be called. I think they're using `xpc_connection_cancel` as a kind of "fatal error".

Here's the transcript from the relevant part of the video, btw:

https://github.com/ASCIIwwdc/wwdc-session-transcripts/blob/2a081ef69334d043bb2475caf1a93eb4fdcd3889/2013/702.vtt#L1725-L1763

Now that I understand why they were doing this, I'll be changing my usage here.

Sources/SecureXPC/XPCConnectionType.swift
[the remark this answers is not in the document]
amomchilov: Actually, it's used in the client and endpoint files, too. I think it should stay standalone. However, the cases could use better names (they shouldn't be client-specific)

Sources/SecureXPC/Client/XPCClient.swift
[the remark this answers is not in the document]
amomchilov: I've updated this to call `fatalError`. I really don't think it should ever end up being called.

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -126,16 +126,36 @@ public class XPCClient {
...
+            // If it were not set, a crash would occur upon calling xpc_connection_resume.
+            xpc_connection_cancel(event)
+        })
+        xpc_connection_resume(connection)
+
+        switch endpoint.kind {
+        case .xpcServiceClient:
+            return XPCServiceClient(serviceName: endpoint.serviceName, connection: connection)
+        case .machServiceClient:
+            return XPCMachClient(serviceName: endpoint.serviceName!, connection: connection)
```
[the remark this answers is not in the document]
amomchilov: Reworking `XPCConnectionType` into an enum with associated values allow me to use pattern matching here, and remove the force unwrap.

Sources/SecureXPC/Client/XPCClient.swift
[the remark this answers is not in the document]
amomchilov: I still have to update the docs here (and elsewhere, probably), but could yo uplease look over my last comment?

Sources/SecureXPC/Client/XPCMachClient.swift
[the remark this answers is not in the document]
amomchilov: > In regards to debug descriptions, I'd think there'd be advantages to having the subclass do that because each subclass has the most comprehensive set of contextual information.

To clarify, I didn't necessarily mean `CustomDebugStringConvertible` (although we should implement that, too), but the breakdown of the value when displayed in the debugger. Stored properties are nice because they're shown there directly, without needing `po obj` or the like.

Sources/SecureXPC/Server/XPCMachServer.swift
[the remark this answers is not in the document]
amomchilov: I've made the message more clear, LMK what you think.

amomchilov (commented):
I think I've addressed most (all?) of your feedback. Wanna have another look?

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -123,20 +123,33 @@ public class XPCClient {
...
+        XPCMachClient(machServiceName: machServiceName)
+    }
+
+	public static func forEndpoint(_ endpoint: XPCServerEndpoint) -> XPCClient {
+        let connection = xpc_connection_create_from_endpoint(endpoint.endpoint)
+
+        xpc_connection_set_event_handler(connection, { (event: xpc_object_t) in
+            fatalError("Is should be impossible for this connection to recieve an event.")
+        })
+        xpc_connection_resume(connection)
```
amomchilov, line 135: FIXME: we shouldn't resume the client here. We should probably defer it until `getConnection()`.

Sources/SecureXPC/Client/XPCClient.swift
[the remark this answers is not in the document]
jakaplan: > NSXPCConnection and the C APIs both take this approach, where connections are just these magical opaque things with very little exposed API. They happen to print their service name (if known) in their description, but they're still annoying to deal with, because e.g. their name isn't visible from the object view in the debugger.

Hmm, I think we just have a difference of opinion when it comes to API design here and I think Apple took the right approach. As far as I see it, this is why `CustomDebugStringConvertible` exists. Unless I'm mistaken the value of `debugDescription` will be visible in the debugger?

> How will "or return an existing one for an XPCEndpoint" work if the connection isn't passed in?

It'd get passed into the initializer of a concrete subclass.

Sources/SecureXPC/Server/XPCMachServer.swift
```diff
@@ -193,6 +196,21 @@ internal class XPCMachServer: XPCServer {
 		return accept
 	}
 
+    public override var endpoint: XPCServerEndpoint {
+        guard let connection = self.serviceListenerConnection else {
+            fatalError("""
+                An XPCServer for a Mach service can only have its `connection` exported to an `endpoint` after
+                a `connection` has been created by calling `start()`.
```
jakaplan, line 203: Remove the backticks since they'll just literally print as backticks?

jakaplan: The concept of a `connection` isn't part of the public API and I don't believe the fact this is for a Mach service is relevant (is it?) so probably better to say something like:
> An XPCServer's endpoint can only be retrieved after start() has been called on it.

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -333,6 +333,14 @@ public class XPCServer {
 	internal func acceptMessage(connection: xpc_connection_t, message: xpc_object_t) -> Bool {
 		fatalError("Abstract Method")
 	}
+
+    public var serviceName: String? {
```
jakaplan, line 337: Looks like `XPCMachServer` doesn't override this.

Separately is this useful for an API user to know? Excluding anonymous servers there's likely to only ever be one server running, and anonymous servers don't have a service name.

Sources/SecureXPC/Server/XPCServiceServer.swift
```diff
@@ -40,9 +40,20 @@ internal class XPCServiceServer: XPCServer {
 
         return (uint32ToString(packageType.bigEndian), uint32ToString(packageCreator.bigEndian))
     }
-    
+
+    private lazy var xpcServiceName: String = {
+        guard let bundleID = Bundle.main.bundleIdentifier else {
+            fatalError("XPC Services need to have a bundle identifier set.")
```
jakaplan, line 46: Hmm, instead of taking this fatalError approach, what if it was determined during the `_forThisXPCService()` function and if it's not present it throws something like `XPCError.misconfiguredXPCService`? That'd make it more consistent with behavior for the already thrown `XPCError.notXPCService` error.

Sources/SecureXPC/Server/XPCServiceServer.swift
```diff
@@ -55,4 +66,23 @@ internal class XPCServiceServer: XPCServer {
...
+
+    public override var serviceName: String? {
+        xpcServiceName
+    }
+
+    public override var endpoint: XPCServerEndpoint {
+        guard let connection = self.connection else {
+            fatalError("""
+                An XPCServer for an XPC service can only have its `connection` exported to an `endpoint` after
+                a `connection` has been created by calling `start()`.
```
jakaplan, line 78: I think this can have the same message used for `XPCMachService`, something like:
> An XPCServer's endpoint can only be retrieved after start() has been called on it.

Sources/SecureXPC/XPCServiceDescriptor.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  XPCServiceDescriptor.swift
+//  
+//
+//  Created by Alexander Momchilov on 2021-11-28.
+//
+
+import Foundation
+
+public enum XPCServiceDescriptor {
```
jakaplan, line 10: Does this need to be `public`? It's usage in `XPCServerEndpoint` is internal only.

jakaplan (changes requested):

Sources/SecureXPC/Server/XPCMachServer.swift
amomchilov: > (is it?)

I don't think that's something that could be reasonably withheld. This framework doesn't exist in isolation: I'd bet almost everyone looking to use it will be cross-referencing against the body of existing literature on XPC, where the existence of the connection is a core concept.

Try to hide the notion of a connection is also a leaky abstraction. Its existence leaks through the errors (e.g. `.connectionInvalidated`), and it's explicitly mentioned in docc docs, in several places. In addition, the message latency is a side channel that hints towards the existing of a connection, and its currency status.

Sources/SecureXPC/Server/XPCServer.swift
amomchilov: I use it in my logger in my project.

Sources/SecureXPC/Server/XPCServiceServer.swift
amomchilov: Great idea.

Sources/SecureXPC/Server/XPCServiceServer.swift
amomchilov: Deduping against https://github.com/trilemma-dev/SecureXPC/pull/23#discussion_r762503640

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -249,6 +262,11 @@ public class XPCClient {
 
     // MARK: Abstract methods
 
+
+    internal var serviceName: String? {
```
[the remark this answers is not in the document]
amomchilov: Ah, it's `public` on my repo. I forgot to upstream that tweak. Fixed.

Sources/SecureXPC/Server/XPCMachServer.swift
jakaplan: I should've been more explicit in my feedback. As written you have `connection` which to me made it sound like that was a property that was actually exposed to the end user. Users never directly create connections or have access to one so depending on how it's worded it could be confusing and just be easier to say it needs to be started, but I have no issue with a connection being mentioned.

In terms of it being for a Mach service, since all of the servers will need to be started it's probably just not very helpful to add the detail that a server for a Mach service needs to be started. This isn't about withholding anything, just that it doesn't seem relevant and therefore may cause the reader to read something into it that isn't applicable.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Is logging that any more useful than logging the bundle id?

Sources/SecureXPC/Server/XPCMachServer.swift
amomchilov: Oh I see, that makes sense.

Sources/SecureXPC/Server/XPCServer.swift
amomchilov: Nah, but that's kind of an implementation detail. The "service name" is what I'm looking to log, and the bundle ID just happens to be the source for it at the moment.

amomchilov:
@jakaplan I think I've addressed almost all the comments. Do you feel comfortable merging as-is and iterating on the remaining stuff? Stale PRs become increasingly hard to work on with time

Sources/SecureXPC/Server/XPCMachServer.swift
```diff
@@ -197,6 +200,18 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
 		return accept
 	}
 
+    public override var endpoint: XPCServerEndpoint {
+        guard let connection = self.serviceListenerConnection else {
+            fatalError("An XPCServer's endpoint can only be retrieved after start() has been called on it.")
+        }
+
+        let endpoint = xpc_endpoint_create(connection)
```
jakaplan, line 208: I'm assuming this actually works? I don't have any good way to test it, but just realized this is rather unintuitive because the connection being serialized into an endpoint here is the one that accepts new connections, not one which can be used to receive events.

Sources/SecureXPC/Server/XPCServiceServer.swift
```diff
@@ -40,9 +44,18 @@ internal class XPCServiceServer: XPCServer {
 
         return (uint32ToString(packageType.bigEndian), uint32ToString(packageCreator.bigEndian))
     }
-    
+
+    private lazy var xpcServiceName: String = {
+        // This is safe to unwrap because it was already checked in ``_forThisXPCService``.
+        Bundle.main.bundleIdentifier!
+    }()
```
jakaplan, line 51: Is this indirection necessary? My understanding is was, quoting from Swift.org, "A lazy stored property is a property whose initial value isn’t calculated until the first time it’s used." If that's true then I think the following ought to be equivalent:
```swift
private lazy var xpcServiceName: String = Bundle.main.bundleIdentifier!
```

Sources/SecureXPC/Server/XPCServiceServer.swift
```diff
@@ -55,4 +68,20 @@ internal class XPCServiceServer: XPCServer {
...
 		true
 	}
+
+    public override var serviceName: String? {
+        xpcServiceName
+    }
+
+    public override var endpoint: XPCServerEndpoint {
+        guard let connection = self.connection else {
+            fatalError("An XPCServer's endpoint can only be retrieved after start() has been called on it.")
```
jakaplan, line 78: `startAndBlock()`?

jakaplan:
> @jakaplan I think I've addressed almost all the comments. Do you feel comfortable merging as-is and iterating on the remaining stuff? Stale PRs become increasingly hard to work on with time

Yes, pending an answer to whether the endpoint for `XPCMachServer` actually works. The other two comments I left were quite minor.

Sources/SecureXPC/Server/XPCServiceServer.swift
amomchilov: Ever since the `guard` was removed from here, no, it's not necessary anymore.

Sources/SecureXPC/Server/XPCMachServer.swift
amomchilov: IIRC, it's "smart" and doesn't encode the service listener connection itself. When a connection it created from this, it's able to talk to the mach service.

However, this isn't something I can test right now, either

jakaplan: Ah, what's stopping you from trying this out? I assume it doesn't actually need to be sent across processes in order to be tested (although normally not doing so would be pointless).

amomchilov: How do you think we should proceed? I'm moving to a new house soon, so I won't be able to build out a test bed for this any time soon. I see two options:

1) Yolo ship it, follow up on it later (with an issue to track it)
2) Fatal-error this code path to say that it's untested, and treat it as effectively not-yet-implemented.

jakaplan: Let's go with option 2 for now. I'd volunteer to try testing it out, but between end of year busyness at work and heading off on holiday soon I just won't have time before early January. Can revisit it then.

amomchilov: Opened https://github.com/trilemma-dev/SecureXPC/issues/33

jakaplan (approved):

PR #24, opened 2021-11-28
Add support for anonymous connections

amomchilov:
This PR builds on top of the code introduced in #23. Do not ship until #23 is merged.

Adds a new `XPCAnonymousServer` that can be passed around using an `XPCEndpoint`. `XPCEndpoint` can then be used to create a client that connects to the anonymous service.

### DIFF
--- a/Sources/SecureXPC/Client/XPCAnonymousServiceClient.swift
+++ b/Sources/SecureXPC/Client/XPCAnonymousServiceClient.swift
@@ -1,0 +1,25 @@
+//
+//  XPCAnonymousServiceClient.swift
+//  SecureXPC
+//
+//  Created by Alexander Momchilov on 2021-12-04
+//
+
+import Foundation
+
+/// A concrete implementation of ``XPCClient`` which can communicate with an anonymous XPC service.
+///
+/// In the case of this framework, the Mach service is expected to be represented by an `XPCMachServer`.
+internal class XPCAnonymousServiceClient: XPCClient {
+    override var serviceName: String? { nil }
+
+    // Anonymous service clients *must* be created from an existing connection.
+    init(connection: xpc_connection_t) {
+        super.init(connection: connection)
+    }
+
+    /// Creates and returns a connection for the Mach service represented by this client.
+    internal override func createConnection() -> xpc_connection_t {
+        fatalError("Anonymous XPC connections cannot be restarted.")
+    }
+}
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -135,6 +135,7 @@ public class XPCClient {
         xpc_connection_resume(connection)
 
         switch endpoint.serviceDescriptor {
+        case .anonymous: return XPCAnonymousServiceClient(connection: connection)
         case .xpcService(name: let name): return XPCServiceClient(xpcServiceName: name, connection: connection)
         case .machService(name: let name): return XPCMachClient(machServiceName: name, connection: connection)
         }
--- a/Sources/SecureXPC/Server/XPCAnonymousServer.swift
+++ b/Sources/SecureXPC/Server/XPCAnonymousServer.swift
@@ -26,6 +26,7 @@ internal class XPCAnonymousServer: XPCServer {
 
     internal override func acceptMessage(connection: xpc_connection_t, message: xpc_object_t) -> Bool {
         // Anonymous service connections should only ever passed among trusted parties.
+        // TODO: add support for client security requirements https://github.com/trilemma-dev/SecureXPC/issues/36
         true
     }
 
--- a/Sources/SecureXPC/Server/XPCAnonymousServer.swift
+++ b/Sources/SecureXPC/Server/XPCAnonymousServer.swift
@@ -1,0 +1,49 @@
+//
+//  XPCAnonymousServer.swift
+//  
+//
+//  Created by Alexander Momchilov on 2021-11-28.
+//
+
+import Foundation
+
+internal class XPCAnonymousServer: XPCServer {
+    private let anonymousListenerConnection: xpc_connection_t
+
+    internal override init() {
+        self.anonymousListenerConnection = xpc_connection_create(nil, nil)
+        super.init()
+
+        // Start listener for the new anonymous connection, all received events should be for incoming client connections
+         xpc_connection_set_event_handler(anonymousListenerConnection, { newClientConnection in
+             // Listen for events (messages or errors) coming from this connection
+             xpc_connection_set_event_handler(newClientConnection, { event in
+                 self.handleEvent(connection: newClientConnection, event: event)
+             })
+             xpc_connection_resume(newClientConnection)
+         })
+    }
+
+    internal override func acceptMessage(connection: xpc_connection_t, message: xpc_object_t) -> Bool {
+        // Anonymous service connections should only ever passed among trusted parties.
+        true
+    }
+
+    /// Begins processing requests received by this XPC server and never returns.
+    public override func startAndBlock() -> Never {
+        fatalError("startAndBlock() is not supported for anonymous connections. Use start() instead.")
+    }
+
+    public override var endpoint: XPCServerEndpoint {
+        XPCServerEndpoint(
+            serviceDescriptor: .anonymous,
+            endpoint: xpc_endpoint_create(self.anonymousListenerConnection)
+        )
+    }
+}
+
+extension XPCAnonymousServer: NonBlockingStartable {
+    public func start() {
+        xpc_connection_resume(self.anonymousListenerConnection)
+    }
+}
--- a/Sources/SecureXPC/Server/XPCServer.swift
+++ b/Sources/SecureXPC/Server/XPCServer.swift
@@ -111,6 +111,10 @@ public class XPCServer {
     public static func forThisXPCService() throws -> XPCServer {
         try XPCServiceServer._forThisXPCService()
     }
+
+    public static func makeAnonymousService() -> XPCServer & NonBlockingStartable {
+        XPCAnonymousServer()
+    }
     
     /// Provides a server for this helper tool if it was installed with
     /// [`SMJobBless`](https://developer.apple.com/documentation/servicemanagement/1431078-smjobbless).
--- a/Sources/SecureXPC/Server/XPCServer.swift
+++ b/Sources/SecureXPC/Server/XPCServer.swift
@@ -112,7 +112,7 @@ public class XPCServer {
         try XPCServiceServer._forThisXPCService()
     }
 
-    public static func makeAnonymousService() -> XPCServer & NonBlockingStartable {
+    internal static func makeAnonymousService() -> XPCServer & NonBlockingStartable {
         XPCAnonymousServer()
     }
     
--- a/Sources/SecureXPC/XPCServiceDescriptor.swift
+++ b/Sources/SecureXPC/XPCServiceDescriptor.swift
@@ -8,6 +8,7 @@
 import Foundation
 
 internal enum XPCServiceDescriptor {
+    case anonymous
     case xpcService(name: String)
     case machService(name: String)
 }


Sources/SecureXPC/Server/XPCAnonymousServer.swift
```diff
@@ -0,0 +1,44 @@
...
+import Foundation
+
+internal class XPCAnonymousServer: XPCServer {
+    private let anonymousListenerConnection: xpc_connection_t
+
+    internal override init() {
+        self.anonymousListenerConnection = xpc_connection_create(nil, nil)
+        super.init()
+
+        // Start listener for the new anonymous connection, all received events should be for incoming client connections
```
jakaplan, line 17: Why is this happening in the initializer instead of in `start()`? Having it occur as part of the `start()` function would make it consistent with `XPCMachServer` and `XPCServiceServer`.

Sources/SecureXPC/Server/XPCAnonymousServer.swift
```diff
@@ -0,0 +1,44 @@
...
+             xpc_connection_set_event_handler(newClientConnection, { event in
+                 self.handleEvent(connection: newClientConnection, event: event)
+             })
+             xpc_connection_resume(newClientConnection)
+         })
+         xpc_connection_resume(anonymousListenerConnection)
+    }
+
+    public override func start() -> Never {
+        fatalError("Anonymous services don't need to be `start()`ed.")
```
jakaplan, line 29: See above comment in regards to some of the initializer functionality being relocated to here.

Sources/SecureXPC/Server/XPCAnonymousServer.swift
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    internal override func acceptMessage(connection: xpc_connection_t, message: xpc_object_t) -> Bool {
+        // Anonymous service connections should only ever passed among trusted parties.
```
jakaplan, line 33: This feels like a security vulnerability waiting to happen. Why not enforce the same sort of requirements that `XPCMachServer` uses? If someone really doesn't want to make the connection secure they can always pass in a requirement that basically anything will satisfy (like `anchor apple generic`).

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -108,6 +108,10 @@ public class XPCServer {
     public static func forThisXPCService() throws -> XPCServer {
         try XPCServiceServer._forThisXPCService()
     }
+
+    public static func newAnonymousService() -> XPCServer {
```
jakaplan, line 112: Based on the [Swift.org API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) we ought to call this `makeAnonymousService()`.

jakaplan, line 112: Based on the [Swift.org API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) we ought to call this `makeAnonymousService()`.

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -137,7 +137,7 @@ public class XPCClient {
         xpc_connection_resume(connection)
 
         switch endpoint.kind {
-        case .xpcServiceClient:
+        case .anonymousClient, .xpcServiceClient:
             return XPCServiceClient(serviceName: endpoint.serviceName, connection: connection)
```
jakaplan, line 141: Hmm, but looking at PR #23 the `XPCServiceClient`'s `createConnection()` function would result in creating a *new* anonymous connection not connecting to this one? I'm guessing this kind of works right now because the connection itself is being passed in and so `createConnection()` isn't actually getting called, but that feels far too brittle of logic.

It's likely clearer to have a distinct `XPCAnonymousClient` class to handle this case separately. Yes it's a bit more code, but it'll be easier to understand what's going on.

Also how does an anonymous client handle when its connection becomes invalid? I don't believe there's any way for it to reestablish the connection. If that's true then it's important all of the `send`/`sendMessage` functions throw a descriptive error when that happens.

jakaplan (changes requested):

Sources/SecureXPC/Server/XPCAnonymousServer.swift
amomchilov: We can’t use `start()` because it returns `Never`, whereas this server doesn’t need to take over the main thread, or anything like that.

jakaplan: I'm not sure I follow about what's wrong with using `start()`? It never returns by choice, not because it must. For example `XPCMachServer` doesn't have to take over the main thread either, it's just convenient for it to do so due to how most helper tools are likely to be written. The onus certainly could be put on the API caller to handle keeping the main thread alive.

This inconsistent behavior is definitely undesirable particularly because only `XPCServer` is the exposed type. That sometimes calling `start()` on an `XPCServer` results in a `fatalError` is too unpredictable. I can absolutely imagine API users relying having functionality that operates on a `XPCServer` where the underlying server type may differ at runtime (I'm actually considering doing this so that I have shared code which can run as either an XPC Service or as a Launch Agent using Mach services).

If there's a good reason to have two different ways to start a server then I think a reasonable approach would be either to create two different start functions or add a `Bool` parameter to specify whether to dispatch the main thread as part of starting. The disadvantage of the latter is its return type can't be `Never` then even if it doesn't actually return.

amomchilov: > I'm not sure I follow about what's wrong with using start()?

See [the usage in the `RoundTripIntegrationTest`](https://github.com/trilemma-dev/SecureXPC/pull/25/commits/dd05992d541827a34e122fd96e4cb5238d1d4bca#diff-1d0119acaf113027616db5ae088a57770b607260d6911dc23b22d29fd2c82c54R15). If we called a `start() -> Never` method, the main thread would block, and it would be game over.

> The onus certainly could be put on the API caller to handle keeping the main thread alive.

What do you mean by "keeping the main thread alive"?

> This inconsistent behavior is definitely undesirable particularly because only `XPCServer` is the exposed type. That sometimes calling `start()` on an `XPCServer` results in a `fatalError` is too unpredictable. 

You're right, that's pretty jank.

> I'm actually considering doing this so that I have shared code which can run as either an XPC Service or as a Launch Agent using Mach services

Funny enough, I was building that too! I wanted to have an easier debugging experience for most things that fully need to run as root.

> If there's a good reason to have two different ways to start a server then I think a reasonable approach would be either to create two different start functions or add a Bool parameter to specify whether to dispatch the main thread as part of starting. The disadvantage of the latter is its return type can't be Never then even if it doesn't actually return.

This design is definitely under-baked. I'll look into it.

Sources/SecureXPC/Client/XPCClient.swift
amomchilov: > Hmm, but looking at PR #23 the XPCServiceClient's createConnection() function would result in creating a new anonymous connection not connecting to this one

MMmm I'm not sure what you mean. When using this `forEndpoint` API, the connection is made on line 130 (from the endpoint), passed to the XPCServiceClient as a param on line 141

> I'm guessing this kind of works right now because the connection itself is being passed in and so `createConnection()` isn't actually getting called, but that feels far too brittle of logic.

That's exactly correct. Why does it feel brittle? It seems pretty reasonable to me: "If you give me my connection, I'll use it, otherwise I'll make my own connection, with blackjack and hookers]"

> It's likely clearer to have a distinct XPCAnonymousClient class to handle this case separately. Yes it's a bit more code, but it'll be easier to understand what's going on.

I've been on the fence about it. I was considering perhaps rolling all the clients together into one class, because unlike the server case, they really don't have that many differences, thus they don't really need polymorhpism that much.

> Also how does an anonymous client handle when its connection becomes invalid?

I haven't gotten that far lol. Happy path only, baby!

Sources/SecureXPC/Server/XPCAnonymousServer.swift
amomchilov: I don't have a good idea for how to design that API yet. Let's ship as-is, and add this on later.

Sources/SecureXPC/Server/XPCAnonymousServer.swift
jakaplan: > What do you mean by "handle keeping the main thread alive"?

I mean have the caller invoke `dispatchMain()` (or equivalent) themselves.

Sources/SecureXPC/Server/XPCAnonymousServer.swift
jakaplan: > I don't have a good idea for how to design that API yet.

One approach would be to refactor a portion of `XPCMachServer` into something like `XPCSecureServer` and then have `XPCMachServer` and `XPCAnonymousServer` inherit from it. Alternatively, the code could be factored out into a helper function or two. Comes down to whether we want to embrace composition or inheritance. In general I feel composition is a better fit for Swift, but we're already making use of inheritance to represent servers.

> Let's ship as-is, and add this on later.

Sorry, but this isn't an area I'm willing to compromise on as is security is the central tenet of this framework. I very much appreciate the added functionality you've brought to this framework, but for me to be comfortable adding new code it always needs to adhere to this.

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: > That's exactly correct. Why does it feel brittle? It seems pretty reasonable to me: "If you give me my connection, I'll use it, otherwise I'll make my own connection, with blackjack and hookers]"

Because there are other cases where a connection may need to be created, such as when it's been invalidated. And this won't work because there's no service name.

amomchilov: I'm afraid I don't follow.

My desired behaviour was this: upon initialization, the client would use whatever connection was provided to it. If it wasn't provided one, it will create its own using `createConnection()`.

If the connection becomes invalidated, it'll be nil-ed out. Upon the next attempt to use  the connection, it'll cause another call to `createConnection()`.

This won't work well for anonymous connections (I need to make its implementation of `createConnection()` to throw an error), but other than that, I don't really see any issue. Could you elaborate on it?

jakaplan: > I'm afraid I don't follow.

The point I was trying to make was exactly this:

> This won't work well for anonymous connections (I need to make its implementation of createConnection() to throw an error), but other than that, I don't really see any issue. Could you elaborate on it?

Sources/SecureXPC/Server/XPCAnonymousServer.swift
amomchilov: I'm rebased on top of the new `NonBlockingStartable` changes. How does it look now?

How do you think `startAndBlock()` should work for `XPCAnonymousServer`? Arbitrarily calling `dispatchMain` or `RunLoop.main.run()` doesn't really "fit"

Sources/SecureXPC/Server/XPCAnonymousServer.swift
amomchilov: I suggest shipping it as-is not because I don't think it's important, but because "out of trunk" development is incredibly costly (in time/effort), and realllllly slows everything down.

How about we ship it as `internal`, and don't expose publically until it has a security-requirements API built in?

> One approach would be to refactor a portion of `XPCMachServer` into something like `XPCSecureServer` and then have `XPCMachServer` and `XPCAnonymousServer` inherit from it.

This is what I had in mind, as well.

It's a massive change though, it would be inappropriate to lump it into a single PR like this

Sources/SecureXPC/Client/XPCClient.swift
amomchilov: How do you think we should design this?

We could have a state-machine that remembers if the anon connection was interrupted/invalidated, and all subsequent send attempts will call reply block with an error, or will call the error handler (after #17)

Similarly, can "invalidated" connection ever be reconnected? (for non-anon servers) If not, what should we do in that case? It should probably be similar to what we choose to do here.

Sources/SecureXPC/Server/XPCAnonymousServer.swift
amomchilov: Hmmm perhaps a compositional approach of having a `MessageAuthenticator` protocol might be preferable. I'll play around with various approaches.

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: > Similarly, can "invalidated" connection ever be reconnected (for non-anon servers)?

The semantics of this matter a great deal. The literal answer to your exact question is "no", but I don't believe it's quite the right question to be asked.

The key to this is that `XPCClient` is not a Swift friendly wrapper around an `xpc_connection_t`. The client represents an entry point to communicate with an `XPCServer`. Under the hood any number of `xpc_connection_t` instances may over the lifetime of the client be used in order to communicate with the server. 

The way an `XPCClient` whose connection was invalidated can reestablish communication with an `XPCServer` is by simply creating another connection. In most cases what probably happened in order for the connection to have become invalidated is the process containing the server terminated and creating a new connection resulted in `launchd` relaunching the process, but that's effectively an implementation detail of XPC.

> If not, what should we do in that case? It should probably be similar to what we choose to do here.

We shouldn't do the same thing because for non-anonymous cases a new connection can be established.

> How do you think we should design the behaviour of disconnected clients (of anonymous services)?

I believe what would make sense is:
 - The initializer for `XPCAnonymousClient` always passes a non-nil `xpc_connection_t` instance in its super initializer call
     - This means the client starts with a `connection` and so calls to `getConnection()` don't result in a call to `createConnection()` unless there's been a reason the connection was nilled out, such as it being invalid
 - In `XPCServer` the function `createConnection()` now `throws`, which in turn means `getConnection()` and `sendWithReply(...)` also need to `throws`
     - In turn this means that all of the `send(...)/sendMessage(...)` functions need to `throws` which they already do so that's not an issue 
 - In `XPCAnonymousClient` the `createConnection()` implementation always throws an error, something like `connectionCannotBeReestablished`

Sources/SecureXPC/Server/XPCAnonymousServer.swift
jakaplan: > How about we ship it as internal, and don't expose publically until it has a security-requirements API built in?

That works for me.

> Hmmm perhaps a compositional approach of having a `MessageAuthenticator` protocol might be preferable. I'll play around with various approaches.

Not sure there's much of a benefit to having a protocol for this if it's all purely internal to the framework? But happy to take a look whatever you come up with.

Sources/SecureXPC/Server/XPCAnonymousServer.swift
jakaplan: I think ideally we'd want to do is block the calling thread, not the main thread (unless that happens to be the calling thread, do you know if there's a robust way to do that? (This makes me also wonder what happens when `dispatchMain()` is called from not the main thread, does it genuinely not return *and* block the main thread?)

If that's approach isn't feasible. We can remove `startAndBlock()` from `XPCServer` and then introduce a `BlockingStartable` interface as well. I rather dislike this approach though as it makes the common case for the non-anonymous servers (which are far more important to the API) a bit more complex to understand.

amomchilov: (This makes me also wonder what happens when `dispatchMain()` is called from not the main thread, does it genuinely not return and block the main thread?)

You just get a crash:

```
Crashed Thread:        9

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    Namespace SIGNAL, Code 4 Illegal instruction: 4
Terminating Process:   exc handler [6395]

Application Specific Information:
BUG IN CLIENT OF LIBDISPATCH: dispatch_main() must be called on the main thread
Performing @selector(echo:) from sender NSButton 0x7f8caf7699d0
```

> I think ideally what we'd want to do is block the calling thread

This would fit the "AndBlock" semantics of the method, but ultimately, it's useless. There's almost never a good reason to be blocking a thread like that.

Sources/SecureXPC/Server/XPCAnonymousServer.swift
amomchilov: Tracking this ins #36

I've updated the visibility, so external API users won't be able to create anonymous services.

Sources/SecureXPC/Client/XPCClient.swift
amomchilov: > Under the hood any number of xpc_connection_t instances may over the lifetime of the client be used in order to communicate with the server.

This is why I think it *is* the right question. As far as I remember, the "invalidation" error occurs when e.g. the service name is wrong, where re-connection attempts are guarenteed to fail. That what I was asking about.

> In most cases what probably happened in order for the connection to have become invalidated is the process containing the server terminated and creating a new connection resulted in launchd relaunching the process, but that's effectively an implementation detail of XPC.

Are you sure this isn't "interrupted"?

> In `XPCAnonymousClient` the `createConnection()` implementation always throws an error, something like `connectionCannotBeReestablished`

Sounds good to me. Tracking this in #37

jakaplan: > Are you sure this isn't "interrupted"?

Oh damn, you're right. Just tested out this exact scenario. Sorry about that.

> This is why I think it is the right question. As far as I remember, the "invalidation" error occurs when e.g. the service name is wrong, where re-connection attempts are guarenteed to fail. That what I was asking about.

Reconnection is not guaranteed to fail on retry and I can rather easily imagine how an API user would make use of the initial failure:
1. create a client
2. try to send a message to the server, get back a `connectionInvalid` error
3. go actually install the blessed helper tool
4. try to send a message again, this should succeed

Sources/SecureXPC/Server/XPCAnonymousServer.swift
jakaplan: > You just get a crash

Ah, yeah we should document this. I'll add an issue to do so.

---

On thinking about this more, it seems actually having `startAndBlock()` for anonymous server call `dispatchMain()` would be useful for testing purposes. By testing I don't mean _us_ writing integration tests for SecureXPC, but instead for users of our API to be able to sub out one of the other server types and check that their code actually works the same as it would when running in an external process. Am I wrong in thinking that's useful?

Sources/SecureXPC/Server/XPCAnonymousServer.swift
```diff
@@ -0,0 +1,44 @@
...
+             // Listen for events (messages or errors) coming from this connection
+             xpc_connection_set_event_handler(newClientConnection, { event in
+                 self.handleEvent(connection: newClientConnection, event: event)
+             })
+             xpc_connection_resume(newClientConnection)
+         })
+         xpc_connection_resume(anonymousListenerConnection)
+    }
+
+    public override func start() -> Never {
```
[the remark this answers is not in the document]
jakaplan: At the very least we should move the non-blocking starting logic over to this function while coming to a decision on how `startAndBlock()` should behave.

jakaplan (changes requested):

Sources/SecureXPC/Client/XPCAnonymousServiceClient.swift
```diff
@@ -0,0 +1,25 @@
...
+    override var serviceName: String? { nil }
+
+    // Anonymous service clients *must* be created from an existing connection.
+    init(connection: xpc_connection_t) {
+        super.init(connection: connection)
+    }
+
+    /// Creates and returns a connection for the Mach service represented by this client.
+    internal override func createConnection() -> xpc_connection_t {
+        fatalError("Anonymous XPC connections cannot be restarted.")
```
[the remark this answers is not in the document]
amomchilov: Yes.

Sources/SecureXPC/Server/XPCAnonymousServer.swift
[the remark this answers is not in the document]
amomchilov: It didn't, it inherits the crashing impl. Prevails I should override with a fatal error with a better message.

In any case, this isn't something I'm looking to address in this PR, it'll need more design thought and refinement

Sources/SecureXPC/Client/XPCClient.swift
amomchilov: Oh yeah, that's a good use case.

amomchilov:
What's left to do here (that's in-scope for this PR)?

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -111,6 +111,10 @@ public class XPCServer {
     public static func forThisXPCService() throws -> XPCServer {
         try XPCServiceServer._forThisXPCService()
     }
+
+    public static func makeAnonymousService() -> XPCServer & NonBlockingStartable {
+        XPCAnonymousServer()
```
jakaplan, line 116: `XPCAnonymousServer` doesn't conform to `XPCAnonymousServer` so this won't compile.

jakaplan:
> What's left to do here (that's in-scope for this PR)?

This PR doesn't currently compile:
 - `XPCAnonymousServer` doesn't conform to `NonBlockingStartable` so `XPCServer`'s `makeAnonymousService()` function isn't valid.
 - `XPCAnonymousServer` can't override `start()` as `XPCServer` doesn't have a `start()` function (that function is defined in the `NonBlockingStartable` protocol)

Once the compilation errors are resolved in a reasonable way it can be merged in.

amomchilov:
@jakaplan Ah yes, I saw that, but the fix was part of another branch. I've moved it here, it all builds correctly now.

jakaplan (approved):

PR #25, opened 2021-11-28
Add message-sending integration test

### LINKED ISSUE #22: Add real XPC integration tests

amomchilov:
Resolves #22

This PR builds on top of the code introduced in #24 ([diff](https://github.com/trilemma-dev/SecureXPC/pull/25/commits/dd05992d541827a34e122fd96e4cb5238d1d4bca)). Do not ship until #24 is merged.

Adds an integration test that tests the message sending/receiving logic using an same-process anonymous XPC service.

### DIFF
--- a/Tests/SecureXPCTests/Client & Server/Round-trip Integration Test.swift
+++ b/Tests/SecureXPCTests/Client & Server/Round-trip Integration Test.swift
@@ -1,0 +1,93 @@
+//
+//  Round-trip Integration Test.swift
+//  
+//
+//  Created by Alexander Momchilov on 2021-11-28.
+//
+
+import XCTest
+@testable import SecureXPC
+
+class RoundTripIntegrationTest: XCTestCase {
+    var xpcClient: XPCClient! = nil
+
+    let anonymousServer = XPCServer.makeAnonymousService()
+
+    override func setUp() {
+        let endpoint = anonymousServer.endpoint
+        xpcClient = XPCClient.forEndpoint(endpoint)
+
+        anonymousServer.start()
+    }
+
+    func testSendWithMessageWithReply() throws {
+        let remoteHandlerWasCalled = self.expectation(description: "The remote handler was called")
+        let replyBlockWasCalled = self.expectation(description: "The echo reply was received")
+
+        let echoRoute = XPCRouteWithMessageWithReply("echo", messageType: String.self, replyType: String.self)
+        try anonymousServer.registerRoute(echoRoute) { msg in
+            remoteHandlerWasCalled.fulfill()
+            return "echo: \(msg)"
+        }
+
+        try self.xpcClient.sendMessage("Hello, world!", route: echoRoute) { result in
+            XCTAssertNoThrow {
+                let response = try result.get()
+                XCTAssertEqual(response, "echo: Hello, world!")
+            }
+
+            replyBlockWasCalled.fulfill()
+        }
+
+        self.waitForExpectations(timeout: 1)
+    }
+
+    func testSendWithoutMessageWithReply() throws {
+        let remoteHandlerWasCalled = self.expectation(description: "The remote handler was called")
+        let replyBlockWasCalled = self.expectation(description: "The pong reply was received")
+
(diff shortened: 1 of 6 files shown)


Tests/SecureXPCTests/Client & Server/Round-trip Integration Test.swift
```diff
@@ -0,0 +1,92 @@
...
+//
+
+import XCTest
+import SecureXPC
+
+class RoundTripIntegrationTest: XCTestCase {
+    var xpcClient: XPCClient! = nil
+
+
+    let anonymousServer = XPCServer.newAnonymousService()
```
jakaplan, line 15: So I understand this correctly all tests in this test case reuse the same server, but a new client is created each time. Wouldn't a new server for each test case so routes don't persist between test cases?

jakaplan (changes requested):
This is awesome, thanks for figuring out how to get this testable!

Tests/SecureXPCTests/Client & Server/Round-trip Integration Test.swift
amomchilov: Nope, the service is new, too.

[XUnit testing frameworks](https://en.wikipedia.org/wiki/XUnit) create one instance of the test class per test being run. This is done specifically to avoid test state being shared by instance variables.

amomchilov: Heh, did you just upvote https://stackoverflow.com/a/68655018/3141234 ?

jakaplan: > XUnit testing frameworks create one instance of the test class per test being run. This is done specifically to avoid test state being shared by instance variables.

Ah, I didn't realize that. Good to know!

> Heh, did you just upvote https://stackoverflow.com/a/68655018/3141234 ?

Never seen that post before (I don't typically read `ios` tagged posts), but it's helpful to read! Considering how this works, a bit unclear to me why `setUp()` exists instead of `init()` or similar. But this chain of thought goes well beyond this PR.

jakaplan (approved):

Tests/SecureXPCTests/Client & Server/Round-trip Integration Test.swift
amomchilov: > Never seen that post before

Ah, I saw this upvoted recently, and it's not a popular topic/question/answer, so I figured it might have been you, haha.

> a bit unclear to me why `setUp()` exists instead of `init()` or similar.

Few thoughts:

1. There's two initializers, so it might not be clear which one to override. (Not sure if one is just an "convenience" initializer, though).
2. Objective-C didn't have definite-variable-initialization, so there was no problem with leaving variables uninitialized in the init, and delegating that to a function like `setUp()`
3. The initializers use some pretty behind-the-scenes concepts, that perhaps Apple didn't want to expose into every single consumer.

PR #28, opened 2021-12-04
Exclusively path based routes - closes #20

### LINKED ISSUE #20: Type errors in routes are conflated with missing routes
With the way the server's routes are looked up today, a type error (e.g. a message is sent expecting a reply, but the server doesn't expect it to need a reply) ends up being diagnosed as `routeNotRegistered`.

What do you think about centralizing all route storage into a single table? If we find a route with a different type than expected, then we can emit a more specific "type error".

This also has the (beneficial, IMO) side-effect of making routes unique by solely their names, and not their "kinds" (with/without message, with/without reply)

<hr>

I ran into this when just trying stuff out and copy-pasting a route on the client and server end, at a time when my client and server code didn't have a "shared" package where I could centralize the route definition. With all duplication like that comes the risk of de-synchronization, and indeed, that's what happened.

jakaplan:
Routes are now only distinct based on path components. The expected message and return types are only used to provide detailed error messages.

### DIFF
--- a/Sources/SecureXPC/Server/XPCServer.swift
+++ b/Sources/SecureXPC/Server/XPCServer.swift
@@ -176,49 +176,61 @@ public class XPCServer {
     
     /// Registers a route that has no message and can't receive a reply.
     ///
-    /// If this route has already been registered, calling this function will overwrite the existing registration. Routes are unique based on their paths.
-    ///
     /// - Parameters:
     ///   - route: A route that has no message and can't receive a reply.
     ///   - handler: Will be called when the server receives an incoming request for this route if the request is accepted.
+    /// - Throws: If this route has already been registered.
     public func registerRoute(_ route: XPCRouteWithoutMessageWithoutReply,
-                              handler: @escaping () throws -> Void) {
+                              handler: @escaping () throws -> Void) throws {
+        if self.routes.keys.contains(route.route) {
+            throw XPCError.routeAlreadyRegistered(route.route.pathComponents)
+        }
+        
         self.routes[route.route] = ConstrainedXPCHandlerWithoutMessageWithoutReply(handler: handler)
     }
     
     /// Registers a route that has a message and can't receive a reply.
     ///
-    /// If this route has already been registered, calling this function will overwrite the existing registration. Routes are unique based on their paths.
-    ///
     /// - Parameters:
     ///   - route: A route that has a message and can't receive a reply.
     ///   - handler: Will be called when the server receives an incoming request for this route if the request is accepted.
+    /// - Throws: If this route has already been registered.
     public func registerRoute<M: Decodable>(_ route: XPCRouteWithMessageWithoutReply<M>,
-                                            handler: @escaping (M) throws -> Void) {
+                                            handler: @escaping (M) throws -> Void) throws {
+        if self.routes.keys.contains(route.route) {
+            throw XPCError.routeAlreadyRegistered(route.route.pathComponents)
+        }
+        
         self.routes[route.route] = ConstrainedXPCHandlerWithMessageWithoutReply(handler: handler)
     }
     
     /// Registers a route that has no message and expects a reply.
     ///
-    /// If this route has already been registered, calling this function will overwrite the existing registration. Routes are unique based on their paths.
-    ///
     /// - Parameters:
     ///   - route: A route that has no message and expects a reply.
     ///   - handler: Will be called when the server receives an incoming request for this route if the request is accepted.
+    /// - Throws: If this route has already been registered.
     public func registerRoute<R: Decodable>(_ route: XPCRouteWithoutMessageWithReply<R>,
-                                            handler: @escaping () throws -> R) {
+                                            handler: @escaping () throws -> R) throws {
+        if self.routes.keys.contains(route.route) {
+            throw XPCError.routeAlreadyRegistered(route.route.pathComponents)
+        }
+        
         self.routes[route.route] = ConstrainedXPCHandlerWithoutMessageWithReply(handler: handler)
     }
     
     /// Registers a route that has a message and expects a reply.
     ///
-    /// If this route has already been registered, calling this function will overwrite the existing registration. Routes are unique based on their paths.
-    ///
     /// - Parameters:
     ///   - route: A route that has a message and expects a reply.
     ///   - handler: Will be called when the server receives an incoming request for this route if the request is accepted.
+    /// - Throws: If this route has already been registered.
     public func registerRoute<M: Decodable, R: Encodable>(_ route: XPCRouteWithMessageWithReply<M, R>,
-                                                          handler: @escaping (M) throws -> R) {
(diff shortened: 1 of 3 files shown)


jakaplan:
@amomchilov here's my first attempt at creating a PR instead of directly committing to the repo. Beyond giving me feedback on the actual code, please let men now if I've done anything wrong or suboptimal in how I've configured this PR.

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -340,54 +310,88 @@ public class XPCServer {
...
+fileprivate protocol XPCHandler {
+    func handle(request: Request, reply: inout xpc_object_t?) throws
 }
 
-fileprivate struct ConstrainedXPCHandlerWithoutMessageWithoutReply: XPCHandlerWithoutMessageWithoutReply {
+fileprivate struct ConstrainedXPCHandlerWithoutMessageWithoutReply: XPCHandler {
     let handler: () throws -> Void
     
-    func handle() throws {
+    func handle(request: Request, reply: inout xpc_object_t?) throws {
```
amomchilov, line 320: The approach you took here causes some duplication, but it has an opportunity to make more specific error messages.

If you're curious to compare notes, checkout what I came up with here: https://github.com/trilemma-dev/SecureXPC/commit/96d9c92430592a21f748d8767b58536a7ecd1ab9

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -340,54 +310,88 @@ public class XPCServer {
...
+            throw XPCError.routeMismatch("Incoming request for route \(request.route.pathComponents) contained a " +
+                                         "message of type \(String(describing: request.route.messageType)), but the " +
+                                         "handler registered with the server does not have a message parameter.")
+        }
+        
+        if var reply = reply {
+            let payload = try self.handler()
+            try Response.encodePayload(payload, intoReply: &reply)
+        } else {
+            throw XPCError.routeMismatch("Incoming request for route \(request.route.pathComponents) does not expect " +
```
amomchilov, line 370: `guard`

amomchilov (approved):
Looks good, it's quite similar to what I had in mind.

I would suggest running this against the integration test I wrote (this is what I wrote it for, lol)

> please let me know if I've done anything wrong or suboptimal in how I've configured this PR.

It's pretty good.

* I would remove "addresses issue #20" from the title, and instead include `closes #20` ([or one of the other keywords that Github recognizes. like `fixes`, `resolves` and their variants](https://docs.github.com/en/enterprise-server@3.0/github/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue).

    When you do that, the pull request would automatically be listed on the issue's "Linked pull requests" list on the right (you can see a sample on #22).

    It would also setup the reciprocal link, of listing the issue in this PR's "Linked issues" list
    
* Rather than mentioning me with @amomchilov, you can directly tag me as a reviewer on the pull request. I haven't been doing that only because I don't have access to change issue/PR metadata (like links, assignees, reviewers and tags).

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -172,61 +172,54 @@ public class XPCServer {
...
                                             handler: @escaping (M) throws -> Void) {
-        let handlerWrapper = ConstrainedXPCHandlerWithMessageWithoutReply(handler: handler)
-        self.routesWithMessageWithoutReply[route.route] = handlerWrapper
+        self.routes[route.route] = ConstrainedXPCHandlerWithMessageWithoutReply(handler: handler)
     }
     
     /// Registers a route that has no message and expects a reply.
     ///
-    /// If this route has already been registered, calling this function will overwrite the existing registration. Routes are unique based on their paths and types.
+    /// If this route has already been registered, calling this function will overwrite the existing registration. Routes are unique based on their paths.
```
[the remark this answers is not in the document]
jakaplan: Good point, it's hard to imagine a valid case for why an API user would intentionally want to do that.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Ah I didn't realize you were working on this, sorry if I've caused wasted effort.

Looking at your approach, they're certainly similar although I prefer the approach I took a bit as it generates more detailed error messages and doesn't have any forced casts.

Sources/SecureXPC/Server/XPCServer.swift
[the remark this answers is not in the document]
jakaplan: I've addressed this, although the solution I have isn't thread safe.

Sources/SecureXPC/Server/XPCServer.swift
amomchilov: > Ah I didn't realize you were working on this, sorry if I've caused wasted effort.

That's kind of on me, I should have communicated that. In general, I have quite a bit more goodies built on top of this framework in my private repo, but it's generally a bunch of hacks and not-yet-shippable code.

I'm working to upstream them as fast as I can haha.

> I prefer the approach I took a bit as it generates more detailed error messages

Indeed, I agree this is nicer. I was apprehensive to do it just because I was too lazy to code out all the cases :D

One critique I'll make however is that this has a "one error at a time" quality to it, which can be frustrating as a user. You run your code, get an error, fix it, run it again, get another error, and so on. It would be nice to have the full error (either your message type is wrong, or your reply type, *or perhaps both*) in one shot.

> doesn't have any forced casts.

If you're new to Swift (though I doubt it, because you write pretty damn good code), I'd caution against always riding the "`try!`/`!`/`as!` is bad" train to its terminal station. I understand the appeal, but there's times (like this one) where I think it can be appropriate to take these escape hatches.

The force casts simply assert "I can't express this type constraint in Swift's type system", but it doesn't mean "I didn't type-check this", because there's type checking *of a different kind*. I literally called my function "typecheck" 😄

jakaplan: > One critique I'll make however is that this has a "one error at a time" quality to it, which can be frustrating as a user. You run your code, get an error, fix it, run it again, get another error, and so on. It would be nice to have the full error (either your message type is wrong, or your reply type, or perhaps both) in one shot.

I'll centralize the error generation and then aggregate the error messages so they can be appended to one another.

> I'd caution against always riding the "try!/!/as! is bad" train to its terminal station. 

I agree with this sentiment, we're likely just at different points along the track. I find it valuable to within reason maximize the amount of issues the compiler can catch for me, so I try to avoid `as!` and `!` in many circumstances for that reason. 

My views on `try!` are quite different. I'm working on a service that needs very high reliability with structured error reporting back to a server in case of failure, so to simply crash the process when something goes wrong would be counterproductive - that's likely influencing the approach I'm taking here.

amomchilov: I see, that makes perfect sense!

PR #30, opened 2021-12-06
Proof of concept for #29

jakaplan:
@amomchilov Giving it more thought, this is the most promising solution I've come up with for #29. Thoughts?


amomchilov (approved):
I think this is a great approach. I don't see any immediate downsides.

PR #31, opened 2021-12-12
`XPCServer` now supports specifying a target queue

jakaplan:
In order to support this, the server now holds an array of weak references to all of the connections that it has received

### DIFF
--- a/Sources/SecureXPC/Server/XPCMachServer.swift
+++ b/Sources/SecureXPC/Server/XPCMachServer.swift
@@ -15,7 +15,7 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
     /// Name of the service.
     private let machServiceName: String
     /// Receives new incoming connections, created once the server is started.
-    private var machService: xpc_connection_t?
+    private var listenerConnection: xpc_connection_t?
     /// The code signing requirements a client must match in order for an incoming request to be handled.
     private let clientRequirements: [SecRequirement]
     
@@ -150,7 +150,7 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
     
     public func start() {
         // Attempts to bind to the Mach service. If this isn't actually a Mach service a EXC_BAD_INSTRUCTION will occur.
-        let machService = machServiceName.withCString { serviceNamePointer in
+        let listenerConnection = machServiceName.withCString { serviceNamePointer in
             return xpc_connection_create_mach_service(
                 serviceNamePointer,
                 self.targetQueue,
@@ -158,7 +158,7 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
         }
         
         // Start listener for the Mach service, all received events should be for incoming connections
-         xpc_connection_set_event_handler(machService, { connection in
+         xpc_connection_set_event_handler(listenerConnection, { connection in
              // Listen for events (messages or errors) coming from this connection
              xpc_connection_set_event_handler(connection, { event in
                  self.handleEvent(connection: connection, event: event)
@@ -166,9 +166,9 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
              self.connections.append(WeakConnection(connection))
              xpc_connection_resume(connection)
          })
-         xpc_connection_resume(machService)
+         xpc_connection_resume(listenerConnection)
         
-        self.machService = machService
+        self.listenerConnection = listenerConnection
     }
     
 	public override func startAndBlock() -> Never {
--- a/Sources/SecureXPC/Server/XPCMachServer.swift
+++ b/Sources/SecureXPC/Server/XPCMachServer.swift
@@ -163,7 +163,7 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
              xpc_connection_set_event_handler(connection, { event in
                  self.handleEvent(connection: connection, event: event)
              })
-             self.connections.append(Weak<xpc_connection_t>(connection))
+             self.addConnection(connection)
              xpc_connection_resume(connection)
          })
          xpc_connection_resume(listenerConnection)
--- a/Sources/SecureXPC/Server/XPCMachServer.swift
+++ b/Sources/SecureXPC/Server/XPCMachServer.swift
@@ -12,7 +12,11 @@ import Foundation
 /// In the case of this framework, the XPC Service is expected to be communicated with by an `XPCMachClient`.
 internal class XPCMachServer: XPCServer, NonBlockingStartable {
     
+    /// Name of the service.
     private let machServiceName: String
+    /// Receives new incoming connections, created once the server is started.
+    private var machService: xpc_connection_t?
+    /// The code signing requirements a client must match in order for an incoming request to be handled.
     private let clientRequirements: [SecRequirement]
     
     /// This should only ever be called from `getXPCMachServer(...)` so that client requirement invariants are upheld.
@@ -149,7 +153,7 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
         let machService = machServiceName.withCString { serviceNamePointer in
             return xpc_connection_create_mach_service(
                 serviceNamePointer,
-                nil, // targetq: DispatchQueue, defaults to using DISPATCH_TARGET_QUEUE_DEFAULT
+                self.targetQueue,
                 UInt64(XPC_CONNECTION_MACH_SERVICE_LISTENER))
         }
         
@@ -159,9 +163,12 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
              xpc_connection_set_event_handler(connection, { event in
                  self.handleEvent(connection: connection, event: event)
              })
+             self.connections.append(WeakConnection(connection))
              xpc_connection_resume(connection)
          })
          xpc_connection_resume(machService)
+        
+        self.machService = machService
     }
     
 	public override func startAndBlock() -> Never {
--- a/Sources/SecureXPC/Server/XPCMachServer.swift
+++ b/Sources/SecureXPC/Server/XPCMachServer.swift
@@ -163,7 +163,7 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
              xpc_connection_set_event_handler(connection, { event in
                  self.handleEvent(connection: connection, event: event)
              })
-             self.connections.append(WeakConnection(connection))
+             self.connections.append(Weak<xpc_connection_t>(connection))
              xpc_connection_resume(connection)
          })
          xpc_connection_resume(listenerConnection)
--- a/Sources/SecureXPC/Server/XPCMachServer.swift
+++ b/Sources/SecureXPC/Server/XPCMachServer.swift
@@ -159,14 +159,14 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
         self.listenerConnection = listenerConnection
       
         // Start listener for the Mach service, all received events should be for incoming connections
-        xpc_connection_set_event_handler(machService, { connection in
+        xpc_connection_set_event_handler(listenerConnection, { connection in
             // Listen for events (messages or errors) coming from this connection
             xpc_connection_set_event_handler(connection, { event in
                 self.handleEvent(connection: connection, event: event)
             })
             xpc_connection_resume(connection)
         })
-        xpc_connection_resume(machService)
+        xpc_connection_resume(listenerConnection)
     }
     
 	public override func startAndBlock() -> Never {
@@ -209,7 +209,7 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
             See https://github.com/trilemma-dev/SecureXPC/issues/33
         """)
 
-        guard let connection = self.serviceListenerConnection else {
+        guard let connection = self.listenerConnection else {
             fatalError("An XPCServer's endpoint can only be retrieved after start() has been called on it.")
         }
 
--- a/Sources/SecureXPC/Server/XPCServer.swift
+++ b/Sources/SecureXPC/Server/XPCServer.swift
@@ -285,10 +285,12 @@ public class XPCServer {
         // The finalizer is called when the connection's retain count has reached zero, so now we need to remove the
         // wrapper from the containing connections array
         xpc_connection_set_finalizer_f(connection, { opaqueWeakConnection in
-            if let opaqueWeakConnection = opaqueWeakConnection {
-                let weakConnection = Unmanaged<WeakConnection>.fromOpaque(opaqueWeakConnection).takeUnretainedValue()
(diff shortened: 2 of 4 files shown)


Sources/SecureXPC/Server/XPCMachServer.swift
```diff
@@ -12,7 +12,11 @@ import Foundation
 /// In the case of this framework, the XPC Service is expected to be communicated with by an `XPCMachClient`.
 internal class XPCMachServer: XPCServer, NonBlockingStartable {
     
+    /// Name of the service.
     private let machServiceName: String
+    /// Receives new incoming connections, created once the server is started.
+    private var machService: xpc_connection_t?
```
jakaplan, line 18: This isn't actually necessary for supporting a target queue. It resolves an issue with the non-blocking start I discovered while working on this PR.

jakaplan:
@amomchilov when you have a chance, would appreciate you taking a look. Is intended to be a very simple change, but there may be opportunities for improvement (such as is there a better way to have an array of weak references?)

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -177,6 +177,36 @@ public class XPCServer {
     // Routes
     private var routes = [XPCRoute : XPCHandler]()
     
+    /// Array of weak references to connections, used to update their dispatch queues
+    internal var connections = [WeakConnection]()
+    
+    /// Weak reference wrapper around an `xpc_connection_t`
+    internal class WeakConnection {
+        weak var connection: xpc_connection_t?
```
amomchilov, line 185: The idiomatic approach to this is to make a generic `Weak<T>` https://stackoverflow.com/a/24128121/3141234

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -177,6 +177,36 @@ public class XPCServer {
...
+    
+    /// The queue used to run the handlers associated with registered routes.
+    ///
+    /// Applying the target queue is asynchronous and non-preemptive and therefore will not interrupt the execution of an already-running handler. The queue
+    /// returned from reading this property will always be the one most recently set even if it is not yet the queue used to run handlers for all incoming requests.
+    public var targetQueue: DispatchQueue? {
+        set {
+            connections.compactMap{ $0.connection }.forEach{ xpc_connection_set_target_queue($0, newValue) }
+            _targetQueue = newValue
+        }
```
amomchilov, line 203: What's a use-case for setting the queue once the connection is already started?

I can't think of a motivating use-case personally, so I'm biased to think the complexity around it isn't worth it.

amomchilov: Also, I think the `_targetQueue`/`targetQueue` pair can just be simplified to a single `targetQueue: DispatchQueue? { didSet { ... } }`.

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -177,6 +177,36 @@ public class XPCServer {
...
+        
+        get {
+            _targetQueue
+        }
+    }
```
amomchilov, line 208: The `_targetQueue`/`targetQueue` pair can just be simplified to a single stored property with a `didSet` observer:

Here's a more Swifty/idiomatic spelling:
```suggestion
    public var targetQueue: DispatchQueue? {
        didSet {
            connection
                .compactMap(\.connection)
                .forEach { xpc_connection_set_target_queue($0, newValue) }
        }
    }
```

Sources/SecureXPC/Server/XPCServiceServer.swift
```diff
@@ -43,10 +43,12 @@ internal class XPCServiceServer: XPCServer {
     
 	public override func startAndBlock() -> Never {
 		xpc_main { connection in
+            xpc_connection_set_target_queue(connection, XPCServiceServer.service.targetQueue)
 			// Listen for events (messages or errors) coming from this connection
 			xpc_connection_set_event_handler(connection, { event in
 				XPCServiceServer.service.handleEvent(connection: connection, event: event)
 			})
+            XPCServiceServer.service.connections.append(WeakConnection(connection))
```
amomchilov, line 51: Should the connections be removed from this pool upon invalidation/interruption/inwhatever?

Sources/SecureXPC/Server/XPCMachServer.swift
```diff
@@ -159,9 +163,12 @@ internal class XPCMachServer: XPCServer, NonBlockingStartable {
              xpc_connection_set_event_handler(connection, { event in
                  self.handleEvent(connection: connection, event: event)
              })
+             self.connections.append(WeakConnection(connection))
              xpc_connection_resume(connection)
          })
          xpc_connection_resume(machService)
+        
+        self.machService = machService
```
amomchilov, line 171: Could be a github thing, but I think the indentation is off here (and a few other places). Could you look it over?

Sources/SecureXPC/Server/XPCMachServer.swift
amomchilov, line 18: I think we should used the term-of-art for this, which IIRC is "listener connection". From the `xpc_connection_create(3)` man page:

> If the service is a MachService advertised by launchd(8), then the listener connection for the named service will receive the new connection in its event handler.

[`xpc_connection_create_mach_service(_:_:_:)`](https://developer.apple.com/documentation/xpc/1448783-xpc_connection_create_mach_servi)

> If the XPC_CONNECTION_MACH_SERVICE_LISTENER flag is given to this method, then the connection returned will be a listener connection.****

amomchilov (commented):
Some comments/q, lmk what you think

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: I don't have a motivating use case, and definitely agree it adds a bit of complexity.

What ultimately motivated this was consistency with setting an error handler and registering routes - both of which can be done after starting the server.

I originally tried several different designs, but liked all them less than this one:
1. A `targetQueue` property with documentation that it'd be ignored if the server was already running
2. A `setTargetQueue(...)` function call which threw an error if the server was already started
3. Having `start()` and `startAndBlock()` now optionally take in a target queue

They were all to some degree simpler to implement, but all made the API for an external user more inconsistent and therefore complicated.  Of the three options \# 3 had the fewest gotchas, but was most inconsistent while \# 1 had the most gotchas (as it fails silently) but was most consistent.

Sources/SecureXPC/Server/XPCServiceServer.swift
jakaplan: I believe the connections themselves won't be held as it's only a weak reference? So the array over time would grow to have more weak wrappers that have a `nil` connection reference, which certainly isn't ideal. Hmm, will give this a think of how to do this in a simple manner.

Sources/SecureXPC/Server/XPCMachServer.swift
jakaplan: The indentation is off, I believe this is the tabs/spaces thing going on. I decided to not change the surrounding lines to make it clearer what the non-whitespace changes actually were.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Ah thanks. Actually going to go with `willSet` to match the current design where the value isn't reflected until it's been applied to all of the connections.

Sources/SecureXPC/Server/XPCServer.swift
amomchilov: Yeah, I don't think there's any need to monomorphize the generic type. That's the compiler's job.

Sources/SecureXPC/Server/XPCServer.swift
amomchilov: What do you think about making the target-queue immutable by taking it as a parameter to the initializer.

There would be no need for a `setTargetQueue`, or public `targetQueue` property setter, and thus, no need to throw an error if already started.

jakaplan: > What do you think about making the target-queue immutable by taking it as a parameter to the initializer.

I originally looked into that, but it's not feasible (or at least would be quite complicated). This is because users of the API don't actually call the initializers they just access them via the factory methods. For example `XPCServiceServer` has a singleton instance which is statically initialized.

Sources/SecureXPC/Server/XPCServiceServer.swift
jakaplan: Okay I've figured out how to do this by using `xpc_connection_set_finalizer_f` and confirmed it works as I expected. However, to do this in a simple way I actually expanded onto `WeakConnection` so that it's rather specialized now and holds onto a reference of its containing server.

amomchilov: Oh that's a great idea.

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -177,6 +177,46 @@ public class XPCServer {
...
+            self.server = server
+        }
+        
+        func removeFromContainer() {
+            self.server?.connections.remove(self)
+        }
+        
+        static func == (lhs: XPCServer.WeakConnection, rhs: XPCServer.WeakConnection) -> Bool {
+            lhs.connectionHash == rhs.connectionHash
+        }
```
amomchilov, line 203: Why not delegate to `xpc_equal`?

Sources/SecureXPC/Server/XPCServer.swift
amomchilov: ~Yeah, those factory methods would just pass on the parameters to the initializes. What's the issue?~

> For example `XPCServiceServer` has a singleton instance which is statically initialized.

Ah.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Because `xpc_equal` doesn't accept `nil` values so it would've been a bunch more lines of code to write, but good point about the possibilities of hash collisions. Will fix.

amomchilov: I see two approaches:

```suggestion
        static func == (lhs: XPCServer.WeakConnection, rhs: XPCServer.WeakConnection) -> Bool {
            xpc_equal(lhs.connection ?? xpc_null_create(), rhs.connection ?? xpc_null_create())
        }
```

```suggestion
        static func == (lhs: XPCServer.WeakConnection, rhs: XPCServer.WeakConnection) -> Bool {
            switch (lhs.connection, rhs.connection) {
                case (nil, nil), (nil, _?), (_?, nil): return false
                case (let a, let b): return xpc_equal(a, b)
            }
        }
```

amomchilov: Nevermind on that second approach, I realized that despite one being `nil` compared to another, the values might still need to be considered equal (in order for removal from the array to work)

jakaplan: Wasn't aware switch statements could be used like that. I pushed a version using if/else if/else, want to take a look?

amomchilov: Actually, I think this approach is general (of trying to do equality "after death") is irreparably flawed.

The hashes can't be relied on for inequality because they're insufficient. The insufficiency causes false positives. The false positives can cause false matches against a (different) item in the array. That then fucks everything way up.

I think the connections array needs to be reworked into a dictionary.

Given that there's no "xpc_id()` (from what I've seen) where `(xpc_id(a) == xpc_id(b)) ≣ xpc_equal(a, b)`, I think we would need to use a `UUID`, counter (backed by an atomic int or whatever),, or some other durable identifier we generate for ourselves.

amomchilov: Swift has really awesome pattern matching. Almost all `if`/`else if`/`else` ladders are more customarily be expressed with a `switch`. https://docs.swift.org/swift-book/LanguageGuide/ControlFlow.html#ID127

Though there are some exceptions (e.g. somewhere in the code base there's a ladder that calls `xpc_equal` a bunch of times. There's no good way to convert that to a `switch`)

jakaplan: > I think the connections array needs to be reworked into a dictionary.

> Given that there's nothing like an xpc_id() function (from what I've seen) where (xpc_id(a) == xpc_id(b)) ≣ xpc_equal(a, b), I think we would need to use a UUID, counter (backed by an atomic int or whatever), or some other durable identifier we generate for ourselves as keys to the dict. These would be durable and exact, and will work even after the connection dies.

It's currently a set, not an array. I think a set can continue to work, but the unique identification can be UUID based instead of hashing the connection. That'd suffice, yeah?

amomchilov: Yeah, that can work, if you just implement `Hashable` and `Equatable` solely in terms of that `UUID` (and perhaps makes sense to lump in `Identifiable`, too).

Though IMO a dictionary is more fitting. You just have `[UUID: Weak<Connection>]`. You would no longer need conformances to `Hashable` and `Equatable`.

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -237,6 +283,22 @@ public class XPCServer {
...
+    internal func addConnection(_ connection: xpc_connection_t) {
+        // Keep a weak reference to the connection and this server, setting this as the context on the connection
+        let weakConnection = WeakConnection(connection, server: self)
+        self.connections.insert(weakConnection)
+        xpc_connection_set_context(connection, Unmanaged.passUnretained(weakConnection).toOpaque())
+        
+        // The finalizer is called when the connection's retain count has reached zero, so now we need to remove the
+        // wrapper from the containing connections array
+        xpc_connection_set_finalizer_f(connection, { opaqueWeakConnection in
+            if let opaqueWeakConnection = opaqueWeakConnection {
```
amomchilov, line 295: When can the `opaqueWeeakConnection` be `nil`? As far as I can tell, that should never happen, and this should be made fail loudly (`fatalError`)

jakaplan: It shouldn't ever be, updated.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Went with the `UUID` approach and kept it a set. Using a set felt a bit clearer to me as it's a less leaky abstraction where the rest of the class when interacting with the connections doesn't need to know how to add or iterate over them.

jakaplan:
@amomchilov want to take one more look over it? I believe I've addressed all of your points.

amomchilov (approved):

PR #39, opened 2022-01-06
Factors out XPC server message acceptance logic

### LINKED ISSUE #36: Add support for client security requirements in anonymous services

jakaplan:
Factors out the security logic into a SecureMessageAcceptor struct.

Resolves #36

### DIFF
--- a/Sources/SecureXPC/Server/SecureMessageAcceptor.swift
+++ b/Sources/SecureXPC/Server/SecureMessageAcceptor.swift
@@ -1,0 +1,70 @@
+//
+//  SecureMessageAcceptor.swift
+//  SecureXPC
+//
+//  Created by Josh Kaplan on 2022-01-06
+//
+
+import Foundation
+
+/// Accepts messages which meet the provided code signing requirements.
+///
+/// Uses undocumented functionality prior to macOS 11.
+internal struct SecureMessageAcceptor {
+    /// At least one of these code signing requirements must be met in order for the message to be accepted
+    internal let requirements: [SecRequirement]
+    
+    func acceptMessage(connection: xpc_connection_t, message: xpc_object_t) -> Bool {
+        // Get the code representing the client
+        var code: SecCode?
+        if #available(macOS 11, *) { // publicly documented, but only available since macOS 11
+            SecCodeCreateWithXPCMessage(message, SecCSFlags(), &code)
+        } else { // private undocumented function: xpc_connection_get_audit_token, available on prior versions of macOS
+            if var auditToken = xpc_connection_get_audit_token(connection) {
+                let tokenData = NSData(bytes: &auditToken, length: MemoryLayout.size(ofValue: auditToken))
+                let attributes = [kSecGuestAttributeAudit : tokenData] as NSDictionary
+                SecCodeCopyGuestWithAttributes(nil, attributes, SecCSFlags(), &code)
+            }
+        }
+
+        // Accept message if code is valid and meets any of the client requirements
+        var accept = false
+        if let code = code {
+            for requirement in self.requirements {
+                if SecCodeCheckValidity(code, SecCSFlags(), requirement) == errSecSuccess {
+                    accept = true
+                }
+            }
+        }
+        
+        return accept
+    }
+    
+    /// Wrapper around the private undocumented function `void xpc_connection_get_audit_token(xpc_connection_t, audit_token_t *)`.
+    ///
+    /// The private undocumented function will attempt to be dynamically loaded and then invoked. If no function exists with this name `nil` will be returned. If
+    /// the function does exist, but does not match the expected signature, the process calling this function is expected to crash. However, because this is only
+    /// called on older versions of macOS which are expected to have a stable non-changing API this is very unlikely to occur.
+    ///
+    /// - Parameters:
+    ///   - _:  The connection for which the audit token will be retrieved for.
+    /// - Returns: The audit token or `nil` if the function could not be called.
+    private func xpc_connection_get_audit_token(_ connection: xpc_connection_t) -> audit_token_t? {
+        // Attempt to dynamically load the function
+        guard let handle = dlopen(nil, RTLD_LAZY) else {
+            return nil
+        }
+        defer { dlclose(handle) }
+        guard let sym = dlsym(handle, "xpc_connection_get_audit_token") else {
+            return nil
+        }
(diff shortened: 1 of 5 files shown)


jakaplan:
@amomchilov when you have the time, would appreciate you taking a look. I went with a compositional approach over inheritance as it seemed simpler to me.

Sources/SecureXPC/Server/SecureMessageAcceptor.swift
```diff
@@ -0,0 +1,70 @@
...
+    /// At least one of these code signing requirements must be met in order for the message to be accepted
+    internal let requirements: [SecRequirement]
+    
+    func acceptMessage(connection: xpc_connection_t, message: xpc_object_t) -> Bool {
+        // Get the code representing the client
+        var code: SecCode?
+        if #available(macOS 11, *) { // publicly documented, but only available since macOS 11
+            SecCodeCreateWithXPCMessage(message, SecCSFlags(), &code)
+        } else { // private undocumented function: xpc_connection_get_audit_token, available on prior versions of macOS
+            if var auditToken = xpc_connection_get_audit_token(connection) {
```
amomchilov, line 23: If you add an `else` case, then `code` will be initialized in all code paths, removing the need for it to be optional.

Better yet, you should just flip this with a guard as well: `guard var auditToken = ... else { return false }`

Sources/SecureXPC/Server/SecureMessageAcceptor.swift
```diff
@@ -0,0 +1,70 @@
...
+        var accept = false
+        if let code = code {
+            for requirement in self.requirements {
+                if SecCodeCheckValidity(code, SecCSFlags(), requirement) == errSecSuccess {
+                    accept = true
+                }
+            }
+        }
+        
+        return accept
```
amomchilov, line 40: This `for` loop is approximating `contains(where:)` (but without the short-circuiting behaviour)

```suggestion
        guard let code = code else { return false }
        
        return self.requirements.contains(where: { SecCodeCheckValidity(code, SecCSFlags(), $0) == errSecSuccess }
```

Sources/SecureXPC/Server/SecureMessageAcceptor.swift
```diff
@@ -0,0 +1,70 @@
...
+    private func xpc_connection_get_audit_token(_ connection: xpc_connection_t) -> audit_token_t? {
+        // Attempt to dynamically load the function
+        guard let handle = dlopen(nil, RTLD_LAZY) else {
+            return nil
+        }
+        defer { dlclose(handle) }
+        guard let sym = dlsym(handle, "xpc_connection_get_audit_token") else {
+            return nil
+        }
+        typealias functionSignature = @convention(c) (xpc_connection_t, UnsafeMutablePointer<audit_token_t>) -> Void
```
amomchilov, line 61: This gets called on every message. It's probably worth caching the resulting function, thought that would be best done in a follow up PR.

Sources/SecureXPC/Server/SecureMessageAcceptor.swift
```diff
@@ -0,0 +1,70 @@
...
+//
+//  Created by Josh Kaplan on 2022-01-06
+//
+
+import Foundation
+
+/// Accepts messages which meet the provided code signing requirements.
+///
+/// Uses undocumented functionality prior to macOS 11.
+internal struct SecureMessageAcceptor {
```
amomchilov, line 13: Should this be behind a protocol, so you can make an "always allow" implementation?

jakaplan: I originally had it written that way, but it ended up with quite a bit more code across multiple files so it didn't seem worth the complexity.

Sources/SecureXPC/Server/SecureMessageAcceptor.swift
jakaplan: My understanding is `code` needs to be optional because both `SecCodeCreateWithXPCMessage` and `SecCodeCopyGuestWithAttributes` expect a parameter of type `UnsafeMutablePointer<SecCode?>` which can't be satisfied with a non-optional type. Is there a way to make this work?

Incorporated feedback on second point about guard statement

Sources/SecureXPC/Server/SecureMessageAcceptor.swift
jakaplan: Had the same thought while refactoring this, but didn't want to make such a change as part of this PR.

Sources/SecureXPC/Server/SecureMessageAcceptor.swift
amomchilov: AH I didn't know that was the expected param type. Usually optional promotion happens, but not in the case of using `&` to generate a temporary pointer. You'd have to check, maybe it does do optional promotion in that case too, IDR.

amomchilov:
Meta point: I would suggest:

1. You make the PR title more useful, to make it more clear when it shows up. E.g., this is part of my GitHub inbox:

    Screen Shot 2022-01-09 at 2 06 45 PM

    Perhaps "Factor out `SecureMessageAcceptor`"?
    
2. Use one of the recognized keywords (listed [here](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) in the PR description, to make GitHub automatically link this PR to the issue. E.g. `Resolves #36`.

Sources/SecureXPC/Server/SecureMessageAcceptor.swift
amomchilov: Fair enough. If we need it in the future, we can just do an "extract interface" refactor.

Sources/SecureXPC/Server/SecureMessageAcceptor.swift
jakaplan: Unfortunately it's not allowed. Xcode outputs:
> Cannot convert value of type 'UnsafeMutablePointer<SecCode>' to expected argument type 'UnsafeMutablePointer<SecCode?>'

Sources/SecureXPC/Server/SecureMessageAcceptor.swift
jakaplan: Yup, can absolutely refactor in the future if warranted

PR #40, opened 2022-01-09
Load `xpc_connection_get_audit_token` only once

jakaplan:
Factors out the loading of `xpc_connection_get_audit_token` so it only happens once per process lifetime

Also changes to the function loading to either succeed or raise a `fatalError` as there's no expectation this will ever fail on the pre-11.0 versions of macOS this code runs on.


jakaplan:
@amomchilov as discussed, here's the follow up PR

amomchilov (approved):
Excellent!

PR #41, opened 2022-01-10
Improves `startAndBlock()` documentation

### LINKED ISSUE #38: Document `startAndBlock()` must be called from main thread
`dispatchMain()` will crash if called from any other thread

jakaplan:
Fixes #38


jakaplan:
@amomchilov Wanted to confirm this is what you had in mind workflow wise when committing changes that don't necessarily need review, but that doing it in this way makes it easier to add feedback later?

PR #42, opened 2022-01-10
Move start logic for anonymous server to `start()`

jakaplan:
All tests pass, including the "Round-trip Integration Test" (@amomchilov thanks for creating those!)

This makes starting behavior consistent between the anonymous server and our two other existing server types.


jakaplan:
@amomchilov care to review? It's quite a small PR.

amomchilov (approved):

PR #43, opened 2022-01-11
Throw an error when attempting to reconnect to an anonymous server

### LINKED ISSUE #37: Throw error when attempting to reconnect to an anonymous service
https://github.com/trilemma-dev/SecureXPC/pull/24/commits/05b2a15a62a7cc2fbf1c6c15e1ed9a35021fc548

jakaplan:
Fixes #37

Now when using trying to send a message to a no longer operational anonymous server after the connection has become interrupted, `XPCError.connectionCannotBeReestablished` will be thrown. "Server Termination Integration Test" has been written to validate this behavior.

### DIFF
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -290,7 +290,7 @@ public class XPCClient {
         fatalError("Abstract Property")
     }
 
-    /// Creates and returns a connection for the service represented by this client or throws an error if it's unable to do so.
+    /// Creates and returns a connection for the service represented by this client.
     internal func createConnection() throws -> xpc_connection_t {
         fatalError("Abstract Method")
     }
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -151,6 +151,9 @@ public class XPCClient {
     ///   - serviceName: The name of the XPC service; no validation is performed on this.
     internal init(connection: xpc_connection_t? = nil) {
         self.connection = connection
+        if let connection = connection {
+            xpc_connection_set_event_handler(connection, self.handleConnectionErrors(event:))
+        }
     }
     
     /// Receives the result of an XPC send. The result is either an instance of the reply type on success or an ``XPCError`` on failure.
@@ -163,7 +166,7 @@ public class XPCClient {
     /// - Throws: If unable to encode the route. No error will be thrown if communication with the server fails.
     public func send(route: XPCRouteWithoutMessageWithoutReply) throws {
         let encoded = try Request(route: route.route).dictionary
-        xpc_connection_send_message(getConnection(), encoded)
+        xpc_connection_send_message(try getConnection(), encoded)
     }
     
     /// Sends a message which will not receive a response.
@@ -174,7 +177,7 @@ public class XPCClient {
     /// - Throws: If unable to encode the message or route. No error will be thrown if communication with the server fails.
     public func sendMessage<M: Encodable>(_ message: M, route: XPCRouteWithMessageWithoutReply<M>) throws {
         let encoded = try Request(route: route.route, payload: message).dictionary
-        xpc_connection_send_message(getConnection(), encoded)
+        xpc_connection_send_message(try getConnection(), encoded)
     }
     
     /// Sends with no message and provides the reply as either a message on success or an error on failure.
@@ -186,7 +189,7 @@ public class XPCClient {
     public func send<R: Decodable>(route: XPCRouteWithoutMessageWithReply<R>,
                                    withReply reply: @escaping XPCReplyHandler<R>) throws {
         let encoded = try Request(route: route.route).dictionary
-        sendWithReply(encoded: encoded, withReply: reply)
+        try sendWithReply(encoded: encoded, withReply: reply)
     }
     
     /// Sends a message and provides the reply as either a message on success or an error on failure.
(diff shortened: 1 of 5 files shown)


jakaplan:
@amomchilov for your review when you have time

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -224,41 +227,60 @@ public class XPCClient {
...
+        } else if xpc_equal(event, XPC_ERROR_CONNECTION_INTERRUPTED) {
+            // Apple documentation:
+            //   Will be delivered to the connection’s event handler if the remote service exited. The connection is
+            //   still live even in this case, and resending a message will cause the service to be launched on-demand.
+            //
+            // While Apple's documentation is technically correct, it's misleading in the case of an anonymous
+            // connection where there is no service. Because there is no service, there is nothing to be relaunched
+            // on-demand. The connection might technically still be alive, but resending a message will *not* work.
+            //
+            // By setting the connection to `nil` when there is no service (indicated by no service name), anonymous
```
amomchilov, line 276: > (indicated by no service name), anonymous clients can ...

This is better expressed in code:

```
if self is XPCAnonymousServer {
    self.connection = nil
}
```

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -268,8 +290,8 @@ public class XPCClient {
         fatalError("Abstract Property")
     }
 
-    /// Creates and returns a connection for the service represented by this client.
-    internal func createConnection() -> xpc_connection_t {
+    /// Creates and returns a connection for the service represented by this client or throws an error if it's unable to do so.
```
amomchilov, line 293: Unnecessary addition to the documentation IMO.

"or throws an error if it's unable to do so" is what `throws` already means

Sources/SecureXPC/Server/XPCAnonymousServer.swift
```diff
@@ -33,6 +33,13 @@ internal class XPCAnonymousServer: XPCServer {
             endpoint: xpc_endpoint_create(self.anonymousListenerConnection)
         )
     }
+    
+    // This exists for testing purposes
+    internal func shutdown() {
```
amomchilov, line 38: Perhaps:

```suggestion
    internal func simulateDisconnectionForTesting() {
```

Tests/SecureXPCTests/Client & Server/Server Termination Integration Test.swift
```diff
@@ -0,0 +1,72 @@
...
+                            } catch {
+                                XCTFail("Unexpected error: \(error). \(XPCError.connectionCannotBeReestablished) " +
+                                        "should have been thrown.")
+                            }
+                        default:
+                            XCTFail("Unexpected error: \(error). \(XPCError.connectionInterrupted) should have been " +
+                                    "returned.")
+                    }
+                case .success(_):
+                    XCTFail("No error was returned. \(XPCError.connectionInterrupted) should have been returned.")
```
amomchilov, line 66: Swift's pattern matching is composable, so you can just do:

```suggestion
           switch response {
                case .failure(let .connectionInterrupted):
                    interruptedExpectation.fulfill()
                    
                    // make another call to the server and this should fail with connectionCannotBeReestablished
                    do {
                        try client.sendMessage("3rd message", route: echoRoute) { _ in }
                        XCTFail("No error was thrown. \(XPCError.connectionCannotBeReestablished) should " +
                                "have been thrown.")
                    } catch XPCError.connectionCannotBeReestablished {
                        cannotBeReestablishedExpectation.fulfill()
                    } catch {
                        XCTFail("Unexpected error: \(error). \(XPCError.connectionCannotBeReestablished) " +
                                "should have been thrown.")
                    }
                case .failure(let _):
                    XCTFail("Unexpected error: \(error). \(XPCError.connectionInterrupted) should have been returned.")
                case .success(_):
                    XCTFail("No error was returned. \(XPCError.connectionInterrupted) should have been returned.")
```

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: Hmm yeah I considered this, but felt it was cleaner to not directly reference a subclass in the superclass and instead have it act based on this server knowing it doesn't correspond to a service.

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: Fair enough

Sources/SecureXPC/Server/XPCAnonymousServer.swift
jakaplan: Seems reasonable. Have made this change.

Overall I dislike having test only code in the product itself, but there's a massive complexity jump to actually spinning up a multiprocess integration test to not need this.

Tests/SecureXPCTests/Client & Server/Server Termination Integration Test.swift
jakaplan: Ah interesting, didn't realize that was possible. Incorporated this with a couple tweaks.

jakaplan:
@amomchilov changes ready for your review

Sources/SecureXPC/Server/XPCAnonymousServer.swift
amomchilov: It might be possible to move this into the test module, using `@testable import` to gain access to otherwise inaccessible `internal` APIs.

But if it's using any `private` APIs, then it won't be possible (I'm on mobile right now so it's hard for me to check this)

amomchilov (approved):
Looks good!

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: To be precise, the fact it's anonymous isn't actually the relevant bit. (However, only anonymous clients have this behavior.) The relevant aspect is there is no *service*, there's just the listener connection. (As an aside, for this reason `XPCAnonymousServiceClient` is actually misnamed - I'll put up a PR to rename it to just `XPCAnonymousClient`.) Because there's no service, there's no for an interrupted connection be to reestablished.

I'm going to leave this as-is for now. But also agree that neither what I have now nor what you're proposing is entirely "pure" from a class inheritance point of view.

Sources/SecureXPC/Server/XPCAnonymousServer.swift
jakaplan: Ah yeah it'd need access to `anonymousListenerConnection` which is private. I'll leave it as-is for now, but happy to revisit this.

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: The listener connection is anonymous. My interpretation is it's anonymous in the sense the listener connection is not associated with a named service managed by launchd.

amomchilov: Discussion continued in #48

PR #44, opened 2022-01-11
Fix missing dispatch queue setting

jakaplan:
- Anonymous server connections weren't having the dispatch queue being set at or able to be updated
 - Incoming connections for the Mach server weren't having their dispatch queue set or update

I realized this while working on #43, but wanted to address it separately


jakaplan:
@amomchilov unless I overlooked something this should be a real quick review

amomchilov (approved):
Nice catch.

Adding tests for this would be nice.

jakaplan:
> Adding tests for this would be nice.

Hmm, I'll have to give some thought on how to do this. I'd want to assert that the server handlers are running on threads belonging to the dispatch queue I created?

Not sure how to test that the listener connection is properly using the dispatch queue, as that's quite internal to the servers.

PR #47, opened 2022-01-12
Create anonymous servers that only accept same process messages

jakaplan:
This is likely a common use case for an anonymous server, so this makes it easier to create while preserving security

As part of this PR, this also changes to having a `MessageAcceptor` protocol as previously discussed. It does add a bit of code in some places, but overall I think it's now the cleanest approach with there being three different acceptance "strategies".

### DIFF
--- a/Sources/SecureXPC/Server/XPCAnonymousServer.swift
+++ b/Sources/SecureXPC/Server/XPCAnonymousServer.swift
@@ -10,16 +10,14 @@ import Foundation
 internal class XPCAnonymousServer: XPCServer {
     private let anonymousListenerConnection: xpc_connection_t
     
-    /// Determines if an incoming request can be handled based on the provided client requirements
-    private let messageAcceptor: SecureMessageAcceptor
-
-    internal init(clientRequirements: [SecRequirement]) {
-        self.anonymousListenerConnection = xpc_connection_create(nil, nil)
-        self.messageAcceptor = SecureMessageAcceptor(requirements: clientRequirements)
+    private let _messageAcceptor: MessageAcceptor
+    override internal var messageAcceptor: MessageAcceptor {
+        _messageAcceptor
     }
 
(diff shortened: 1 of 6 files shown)


jakaplan:
@amomchilov I've also updated the tests to make use of this and all of them still pass

Sources/SecureXPC/Server/XPCAnonymousServer.swift
```diff
@@ -10,16 +10,14 @@ import Foundation
...
     
-    /// Determines if an incoming request can be handled based on the provided client requirements
-    private let messageAcceptor: SecureMessageAcceptor
-
-    internal init(clientRequirements: [SecRequirement]) {
-        self.anonymousListenerConnection = xpc_connection_create(nil, nil)
-        self.messageAcceptor = SecureMessageAcceptor(requirements: clientRequirements)
+    private let _messageAcceptor: MessageAcceptor
+    override internal var messageAcceptor: MessageAcceptor {
+        _messageAcceptor
```
amomchilov, line 15: What's the purpose of having one private and one internal var?

amomchilov (approved):
Great clean up, I like it!

Sources/SecureXPC/Server/XPCAnonymousServer.swift
jakaplan: In the case of `XPCMachServer` I need to make use of the fact that the message accept is actually of type `SecureMessageAcceptor` so I could access its `requirements` property.

For the anonymous and XPC Services servers, I don't have such a need. But I though there still needed to be a backing variable? How would you write this code? I'm guessing I lack some knowledge of how Swift properties work.

jakaplan: I'm going to merge in this code, but absolutely can revise this part later once I understand how to better write this code.

PR #48, opened 2022-01-13
Removes references to "service" in conjunction with "anonymous"

jakaplan:
Anonymous connections have no service, which is a key distinguishing factor from XPC Services and Mach services.

### DIFF
--- a/Sources/SecureXPC/Client/XPCAnonymousClient.swift
+++ b/Sources/SecureXPC/Client/XPCAnonymousClient.swift
@@ -1,5 +1,5 @@
 //
-//  XPCAnonymousServiceClient.swift
+//  XPCAnonymousClient.swift
 //  SecureXPC
 //
 //  Created by Alexander Momchilov on 2021-12-04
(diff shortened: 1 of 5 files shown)


jakaplan:
@amomchilov Onboard with this? It's a change to terminology, there's no change to functionality in this PR.

amomchilov:
(Continued from [this thread](https://github.com/trilemma-dev/SecureXPC/pull/43/files#r783542910)

Yes you're correct, but my point was that the name `XPCAnonymousClient` makes it sound like an "an anonymous client" not "the client of an anonymous service".

Perhaps there should be a little blurb in the API doc that clarifies that.

But overall, I think the new name is fine

jakaplan:
> Perhaps there should be a little blurb in the API doc that clarifies that.

Do you mean in the public API doc? `XPCAnonymousClient` isn't part of the public API. From an API user's point of view the way to create a client which talks with an anonymous server is `XPCClient.forEndpoint(...)`.

The internal documentation currently states:
> A concrete implementation of ``XPCClient`` which can communicate with an anonymous XPC listener connection.

Is there some additional documentation you'd find helpful? More than happy to expand upon it; if you haven't already gathered I'm a big fan of documentation.

amomchilov:
> A concrete implementation of XPCClient which can communicate with an anonymous XPC listener connection.

OH I didn't notice that was already. THat's perfect, and exactly what I had in mind. 👍

PR #49, opened 2022-01-13
Fixes a racy connection finalizer issue exposed by `testShutdownServer`

jakaplan:
It turns out occassionally `ServerTerminationIntegrationTest`'s `testShutdownServer` would result in a segfault inside of the finalizer for the connection. I'm pretty sure the issue was that a retained count needs to be taken for the WeakConnection instance (what we don't want to do is add a retain count to the connection itself).

### DIFF
--- a/Sources/SecureXPC/Server/XPCServer.swift
+++ b/Sources/SecureXPC/Server/XPCServer.swift
@@ -291,7 +291,7 @@ public class XPCServer {
         // Keep a weak reference to the connection and this server, setting this as the context on the connection
         let weakConnection = WeakConnection(connection, server: self)
         self.connections.insert(weakConnection)
-        xpc_connection_set_context(connection, Unmanaged.passUnretained(weakConnection).toOpaque())
+        xpc_connection_set_context(connection, Unmanaged.passRetained(weakConnection).toOpaque())
         
         // The finalizer is called when the connection's retain count has reached zero, so now we need to remove the
         // wrapper from the containing connections array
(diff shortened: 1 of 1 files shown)


jakaplan:
@amomchilov I basically tested this by just manually running the integration test over and over again. Before it was segfaulting roughly 20% of the time, after this change I've yet to see a segfault. That's not entirely satisfying, but seems strictly better than not addressing this issue.

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -291,7 +291,7 @@ public class XPCServer {
         // Keep a weak reference to the connection and this server, setting this as the context on the connection
         let weakConnection = WeakConnection(connection, server: self)
         self.connections.insert(weakConnection)
-        xpc_connection_set_context(connection, Unmanaged.passUnretained(weakConnection).toOpaque())
+        xpc_connection_set_context(connection, Unmanaged.passRetained(weakConnection).toOpaque())
```
amomchilov, line 294: Hmmm I didn't think this was necessary, because I thought that the `connections` array would be holding a strong reference.

In retrospect, that was silly, I was acting as if strong refs should be spared. I think passing retained, and taking retained is the correct semantics, but only if the finalizer below is guaranteed to be called,  exactly once.

amomchilov (approved):

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: I'm pretty confident the finalizer can only be called once because the documentation says:
> The function that will be invoked when the connection's retain count has dropped to zero and is being torn down.

PR #50, opened 2022-01-13
Removes `endpoint` from `XPCServer` as it can't be supported by `XPCServicesServer`

### LINKED ISSUE #46: Verify that `endpoint` works for `XPCServiceServer`
For the anonymous and the Mach services server, the connection turned into an endpoint is the listener connection. XPC Services don't have an (exposed) listener connection, so the implementation right now uses the incoming connection passed to `xpc_main` which likely does not work. This issue is to verify whether or not this works.

jakaplan:
Fixes #46

`NonBlockingStartable` has been renamed to `NonBlockingServer` and `endpoint` was added to it. Not sure this is the best name for the protocol, but good enough for putting this up for review.

### DIFF
--- a/Sources/SecureXPC/Server/XPCAnonymousServer.swift
+++ b/Sources/SecureXPC/Server/XPCAnonymousServer.swift
@@ -27,13 +27,6 @@ internal class XPCAnonymousServer: XPCServer {
         fatalError("startAndBlock() is not supported for anonymous connections. Use start() instead.")
     }
 
-    public override var endpoint: XPCServerEndpoint {
-        XPCServerEndpoint(
-            serviceDescriptor: .anonymous,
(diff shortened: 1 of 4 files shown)


jakaplan:
@amomchilov for your review

amomchilov:
re: https://github.com/trilemma-dev/SecureXPC/issues/46#issuecomment-1011788484

I think you should document this finding somewhere in the implementation, because it's not obvious that this shouldn't work.

amomchilov (approved):

jakaplan:
> re: [#46 (comment)](https://github.com/trilemma-dev/SecureXPC/issues/46#issuecomment-1011788484)
> 
> I think you should document this finding somewhere in the implementation, because it's not obvious that this shouldn't work.

Good point, will do!

PR #52, opened 2022-01-14
`XPCAnonymousServer` now supports `startAndBlock()`

jakaplan:
It now exactly mimics the implementation used by `XPCMachServer`. While in most cases this wouldn't be useful and `start()` should be used instead, by providing parity it means API users can test out running a Mach service (except for potentially the root access it has) with identical usage of SecureXPC.


jakaplan:
@amomchilov thoughts? I know we've discussed this in the past. Having given it a bunch more though I think this is the best approach because of its consistency. It's true that in many cases it wouldn't be useful, but on balance I think that's okay.

amomchilov:
Interesting thought. My thoughts on this are purely speculative, I haven't gotten to the stage where I'm testing my Mach service yet.

I'll give some feedback when I get to that point.

amomchilov (approved):

PR #53, opened 2022-01-14
Makes `XPCServerEndpoint` conform to `Codable`

### LINKED ISSUE #12: Add support for encoding/decoding XPC connections
<hr>

_Originally posted by @amomchilov in https://github.com/trilemma-dev/SecureXPC/issues/6#issuecomment-960865476_:

> File descriptors and shared memory are intentionally not supported as they are concepts incompatible with Codable which is designed for indefinite serialization and cross-device transfer. As such choosing Codable for SecureXPC had tradeoffs because XPC does not have the same requirements — it can represent live resources on a single device. However, I decided the tradeoff was worth it because of Swift's excellent built in support for Codable (particularly the compiler automatically generating conformance in many cases). If there was a desire to support file descriptors and shared memory, I can envision ways to add them in the future, although it would mean expanding the API quite considerably.

I don't have an immediate need for passing FDs or shared memory regions in my program yet, but my app *does* need to pass an `xpc_endpoint_t`. My main app is sandboxed, so the only way it can use `SMBless` is by delegating that work to an unsandboxed XPC helper service. Once the installation is done, it opens the XPC connection to the privileged helper service, and passes the endpoint back to the main app.

This is the approach that's recommended in Apple's [EvenBetterAuthorizationSample](https://developer.apple.com/library/archive/samplecode/EvenBetterAuthorizationSample/Listings/Read_Me_About_EvenBetterAuthorizationSample_txt.html#//apple_ref/doc/uid/DTS40013768-Read_Me_About_EvenBetterAuthorizationSample_txt-DontLinkElementID_17), so I think supporting `xpc_endpoint_t` is pretty crucial. So at a minimum, at least *some* level of special-casing needs to happen.

jakaplan:
Fixes #12 

Encoding and decoding "cheats" by using specific functions custom to `XPCEncoderImpl` and `XPCDecoderImpl`. As such if any other encoder or decoder tried to make use of these an error will be thrown.

Added two simple roundtrip tests to verify encoding and decoding works as expected for the anonymous server. However, this is a pretty limited test as it doesn't actually send the endpoint over an XPC connection.

### DIFF
--- a/Sources/SecureXPC/XPC Coders/XPCDecoder/XPCKeyedDecodingContainer.swift
+++ b/Sources/SecureXPC/XPC Coders/XPCDecoder/XPCKeyedDecodingContainer.swift
@@ -165,16 +165,6 @@ internal class XPCKeyedDecodingContainer<K: CodingKey>: KeyedDecodingContainerPr
     // MARK: XPC specific decoding
     
     func decodeEndpoint(forKey key: K) throws -> xpc_endpoint_t {
-        let value = try value(forKey: key)
-        let valueType = xpc_get_type(value)
-        if valueType != XPC_TYPE_ENDPOINT {
-            let message = "Expected type: \(XPC_TYPE_ENDPOINT.description)\nActual type: \(valueType.description)"
-            let context = DecodingError.Context(codingPath: self.codingPath,
-                                                debugDescription: message,
-                                                underlyingError: nil)
-            throw DecodingError.typeMismatch(xpc_object_t.self, context)
-        }
-        
-        return value
+        return try decode(key: key, xpcType: XPC_TYPE_ENDPOINT, transform: {$0})
     }
 }
--- a/Sources/SecureXPC/XPC Coders/XPCDecoder/XPCKeyedDecodingContainer.swift
+++ b/Sources/SecureXPC/XPC Coders/XPCDecoder/XPCKeyedDecodingContainer.swift
@@ -161,4 +161,20 @@ internal class XPCKeyedDecodingContainer<K: CodingKey>: KeyedDecodingContainerPr
 	func superDecoder(forKey key: K) throws -> Decoder {
 		return XPCDecoderImpl(value: try value(forKey: key), codingPath: self.codingPath + [key])
 	}
+    
+    // MARK: XPC specific decoding
+    
+    func decodeEndpoint(forKey key: K) throws -> xpc_endpoint_t {
+        let value = try value(forKey: key)
+        let valueType = xpc_get_type(value)
+        if valueType != XPC_TYPE_ENDPOINT {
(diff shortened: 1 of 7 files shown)


jakaplan:
@amomchilov thoughts? I know you were the one to originally propose doing this so, let me know if this is along the lines of what you had in mind.

jakaplan:
> I'm struggling to review it just from GitHub. I'll try to follow up on this over the weekend. I'll need to actually open this open locally and browse around to regain some context.

No worries, I completely understand. Once this is merged, I think we'll be ready to make the anonymous server public along with updated documentation.

Sources/SecureXPC/XPC Coders/XPCDecoder/XPCKeyedDecodingContainer.swift
```diff
@@ -161,4 +161,20 @@ internal class XPCKeyedDecodingContainer<K: CodingKey>: KeyedDecodingContainerPr
...
+        if valueType != XPC_TYPE_ENDPOINT {
+            let message = "Expected type: \(XPC_TYPE_ENDPOINT.description)\nActual type: \(valueType.description)"
+            let context = DecodingError.Context(codingPath: self.codingPath,
+                                                debugDescription: message,
+                                                underlyingError: nil)
+            throw DecodingError.typeMismatch(xpc_object_t.self, context)
+        }
+        
+        return value
+    }
```
amomchilov, line 179: Can't this just reuse `decode(key:xpcType:transform)` like most of the other methods areound here?

Something like:

```suggestion
    func decodeEndpoint(forKey key: K) throws -> xpc_endpoint_t {
        return try decode(key: key, xpcType: XPC_TYPE_ENDPOINT, transform: \.self)
    }
```

amomchilov (approved):
1 q, but otherwise looks good

Sources/SecureXPC/XPC Coders/XPCDecoder/XPCKeyedDecodingContainer.swift
jakaplan: Huh, yeah not sure why I didn't do that. However, I've never seen the `\.self` syntax before and it doesn't compile. This works as expected:
```swift
func decodeEndpoint(forKey key: K) throws -> xpc_endpoint_t {
    return try decode(key: key, xpcType: XPC_TYPE_ENDPOINT, transform: {$0})
}
```

amomchilov: Huh, weird, I thought that was implemented already: https://github.com/apple/swift-evolution/blob/main/proposals/0227-identity-keypath.md

Issue #54, opened 2022-01-14
Creating an `XPCAnonymousServer` or `XPCMachServer` and never starting it results in XPC API Misuse


jakaplan:
If either of these types of servers are created, but never started, then when they are deallocated the following is thrown:
 > XPC API Misuse: Release of last reference on a suspended connection.

After a bunch of debugging and documentation reading I'm pretty sure what's going on is the following:
- "XPC connections, like dispatch sources, are returned in a suspended state [...] connections must be resumed in order to be safely released. It is a programming error to release a suspended connection."
    -  Source: [xpc_connection_create(...) discussion](https://developer.apple.com/documentation/xpc/1448791-xpc_connection_create#discussion)
- Both `XPCAnonymousServer` and `XPCMachServer` create listener connections as part of their initialization, but do not call `xpc_connection_resume(...)` on them as part of initializing
- (If either of those servers are started, then resume *is* called and this issue is not experienced)
- [`xpc_release`](https://developer.apple.com/documentation/xpc/1505851-xpc_release) is unavailable in Swift, but appears to be called by the Swift runtime (or possibly inserted at compile time?) when the listener connection's retain count reaches zero due to the runtime's builtin reference counting
- When `xpc_release` is called on a listener connection the application/executable/test harness experiences a EXC_BAD_INSTRUCTION (SIGILL) due to the aforementioned: "XPC API Misuse: Release of last reference on a suspended connection."

Reproducing this issue with a test is trivial, just create a new test and the entire function's body should just be:
```swift
XPCServer.makeAnonymous()
```

PR #56, opened 2022-01-14
Proper server deallocation, fixes issue #54

jakaplan:
`XPCAnonymousServer` and `XPCMachServer` now both start their listener connections in their initializers. However, now when incoming connections are received they don't unconditionally configure and start them. If the server has not been started, the connections are added to a pending array. Once the server is started, those pending connections are configured and started.

Summary of changes:
 - Consolidated all client connection to `XPCServer.startClientConnection(...)` which all three subclasses now call
 - Each instance of `XPCAnonymousServer` and `XPCMachServer` now has its own dispatch queue that it uses for both receiving connections as well as starting the server and processing pending connections, this should prevent race conditions. `XPCServer.targetQueue` now exclusively applies to running the handlers associated with the incoming client requests which makes more sense (and is actually what our own documentation says happens).
 - Added a seemingly trivial test which now passes, but without these changes always resulted in a SIGILL.

### DIFF
--- a/Sources/SecureXPC/Server/XPCAnonymousServer.swift
+++ b/Sources/SecureXPC/Server/XPCAnonymousServer.swift
@@ -8,18 +8,34 @@
 import Foundation
 
 internal class XPCAnonymousServer: XPCServer {
+    /// Receives new incoming connections
     private let anonymousListenerConnection: xpc_connection_t
-    
+    /// The dispatch queue used when new connections are being received
+    private let listenerQueue = DispatchQueue(label: String(describing: XPCAnonymousServer.self))
+    /// Whether this server has been started, if not connections are added to pendingConnections
+    private var started = false
+    /// Connections received while the server is not started
+    private var pendingConnections = [xpc_connection_t]()
+    /// Determines if an incoming request will be accepted based on the provided client requirements
     private let _messageAcceptor: MessageAcceptor
     override internal var messageAcceptor: MessageAcceptor {
         _messageAcceptor
     }
-
+    
     internal init(messageAcceptor: MessageAcceptor) {
         self._messageAcceptor = messageAcceptor
-        self.anonymousListenerConnection = xpc_connection_create(nil, nil)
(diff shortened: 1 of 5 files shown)


jakaplan:
@amomchilov for your review

FYI I'll be away for the next three days, probably will have pretty spotty Internet access.

amomchilov:
There's a lot of different changes here, and I'm struggling to understand how they're related.

> If the server has not been started, the connections are added to a pending array. Once the server is started, those pending connections are configured and started.

Could you please elaborate on the motivation for this queuing mechanic?

jakaplan:
> > If the server has not been started, the connections are added to a pending array. Once the server is started, those pending connections are configured and started.
> 
> 
> 
> Could you please elaborate on the motivation for this queuing mechanic?

The listener connection needs to be started in the server's initializer to avoid misuse of the XPC API. This means client connections can now be received by the listener connection. However, we don't want to start receiving events from those client connections yet. By storing the pending connections in an array, then once the server is started they can all be activated (resumed) which will result in processing their events.

amomchilov:
> to avoid misuse of the XPC API

Ah was this because deallocated a connection that wasn't started causes an error?

Makes sense.

> However, we don't want to start receiving events from those client connections yet.

Devil's advocate: why not? What if we rolled the `start`ing functionality into the initializer?

jakaplan:
> Ah was this because deallocated a connection that wasn't started causes an error?

Indeed, I tried to document this in some detail in issue #54. Turns out this misuse actually is documented by Apple.

> > However, we don't want to start receiving events from those client connections yet.
> 
> Devil's advocate: why not? What if we rolled the `start`ing functionality into the initializer?

If we rolled `start`ing functionality into the initializer it would mean we'd have to also rolling registering routes into the initializer as well. Routes need to be registered before the server is started because at least one incoming event will be received immediately upon resuming (aka `start`ing) the connection as that's what cause the connection to be received by the listener connection.

Additionally, this would be rather weird for XPC Services as the only way to start them is in a blocking manner so the initializer itself would never return.

amomchilov (approved):

Issue #19, opened 2021-11-14
Add async variants of `XPCClient.send/sendMessage(...)` functions


jakaplan:
macOS 10.15 and later should support this (at least as of the latest Xcode beta)

jakaplan:
This should be easy to implement a meaningful way now that Xcode 13.2 supports back deploying async/await to macOS 10.15.

PR #57, opened 2022-01-18
Optional completion handler `send` and `sendMessage` without reply

### LINKED ISSUE #17: Add support for handling client errors on reply-less messages

jakaplan:
Fixes #17

All `send` and `sendMessage` functions now take in a `XPCResponseHandler` (renamed from `XPCReplyHandler`). For those that require a reply, the response handler is non-optional. For those that can't have a reply, they can optionally provide a `XPCResponseHandler<Void>` which will get called upon successful completion or be provided an `XPCError` on failure.

In all cases, none of the functions now `throw` - all errors generated are provided to their handlers (or ignored if no handler is provided).

Three existing tests were modified and two new tests have been added.

---

This has been designed in a way which should make it relatively straightforward to address #19.

### DIFF
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -157,87 +157,154 @@ public class XPCClient {
     }
     
     /// Receives the result of an XPC send. The result is either an instance of the reply type on success or an ``XPCError`` on failure.
-    public typealias XPCReplyHandler<R> = (Result<R, XPCError>) -> Void
+    public typealias XPCResponseHandler<R> = (Result<R, XPCError>) -> Void
     
     /// Sends with no message and will not receive a reply.
     ///
     /// - Parameters:
     ///   - route: The server route which will handle this.
-    /// - Throws: If unable to encode the route. No error will be thrown if communication with the server fails.
-    public func send(route: XPCRouteWithoutMessageWithoutReply) throws {
-        let encoded = try Request(route: route.route).dictionary
-        xpc_connection_send_message(try getConnection(), encoded)
+    ///   - onCompletion: An optionally provided function or closure to receive a response upon successful completion or error.
+    public func send(route: XPCRouteWithoutMessageWithoutReply,
+                     onCompletion handler: XPCResponseHandler<Void>?) {
+        if let handler = handler {
+            do {
+                let encoded = try Request(route: route.route).dictionary
+                sendWithResponse(encoded: encoded, withResponse: handler)
+            } catch {
+                handler(.failure(.encodingError(String(describing: error))))
+            }
+        } else {
+            if let encoded = try? Request(route: route.route).dictionary,
+               let connection = try? getConnection() {
+                xpc_connection_send_message(connection, encoded)
+            }
+        }
     }
     
-    /// Sends a message which will not receive a response.
+    /// Sends a message which will not receive a reply.
     ///
     /// - Parameters:
-    ///    - message: Message to be sent.
-    ///    - route: The server route which should handle this message.
-    /// - Throws: If unable to encode the message or route. No error will be thrown if communication with the server fails.
-    public func sendMessage<M: Encodable>(_ message: M, route: XPCRouteWithMessageWithoutReply<M>) throws {
-        let encoded = try Request(route: route.route, payload: message).dictionary
-        xpc_connection_send_message(try getConnection(), encoded)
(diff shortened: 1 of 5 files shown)


jakaplan:
@amomchilov for your review. The significant changes are all in `XPCClient`, although a few other files were lightly touched as well.

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -157,87 +157,153 @@ public class XPCClient {
...
+                                                withResponse handler: @escaping XPCResponseHandler<R>) {
+        // Get the connection or inform the handler of failure and return
+        var connection: xpc_connection_t?
+        do {
+            connection = try getConnection()
+        } catch {
+            if let error = error as? XPCError {
+                handler(.failure(error))
+            } else {
+                handler(.failure(.unknown))
```
amomchilov, line 251: This kind of logic is repeated a few times, perhaps it would be nice to put it in an initializer on `XPCError`

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -157,87 +157,153 @@ public class XPCClient {
...
     
-    /// Does the actual work of sending an XPC message which receives a reply.
-    private func sendWithReply<R: Decodable>(encoded: xpc_object_t,
-                                             withReply reply: @escaping XPCReplyHandler<R>) throws {
-        xpc_connection_send_message_with_reply(try getConnection(), encoded, nil, { xpcResponse in
+    /// Does the actual work of sending an XPC message which receives a response.
+    private func sendWithResponse<R: Decodable>(encoded: xpc_object_t,
+                                                withResponse handler: @escaping XPCResponseHandler<R>) {
+        // Get the connection or inform the handler of failure and return
+        var connection: xpc_connection_t?
```
amomchilov, line 244: I don't think Swift requires this to be mutable or optional, because it's always initialized once in all code paths

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -157,87 +157,153 @@ public class XPCClient {
...
-            self.handleConnectionErrors(event: xpcResponse)
-            reply(result)
-        })
+            self.handleConnectionErrors(event: reply)
+            handler(result)
+        }
     }
-
+    
+    /// Wrapper that handles responses without a payload since `Void` is not `Decodable`
```
amomchilov, line 290: This is a real fly in the ointment, eh?

I got around this in my project by making a separate API whose error handler is `(XPCError?) -> Void`. I think the usability is a *tiny* bit nicer at the call-site, but it does make it less consistent with the other APIs.

amomchilov (approved):
Looks good to me!

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: Heh, was thinking that exactly as I wrote this a few times, but didn't want to make the PR too big. Am planning to do this soon.

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: But it's only initialized in one code path? If `try getConnection()` succeeds it's initialized, but if the `catch` block is entered then it's not initialized. Maybe I'm missing something.

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: Oh hmm, so I originally wrote it up exactly like that but decided against it as it felt less clear because my intention wasn't to be an _error_ handler, but a _completion_ handler. So with this design pattern you can check for success instead of needing to check for the non-presence of an error. They are equivalent, but it just seemed a bit clearer to do it this way.

Sources/SecureXPC/Client/XPCClient.swift
amomchilov: The `catch` path should `return`, then what I said holds true.

Minimal sample:

```swift
func thrower() throws -> Int { 123 }

func demo() {
	let i: Int
	
	do {
		i = try thrower()
	} catch {
		return
	}
	
	print(i)
}

demo()
```

jakaplan: Ah yup, that makes complete sense. Will do.

PR #59, opened 2022-01-19
Add async variants of `XPCClient.send/sendMessage(...)`

### LINKED ISSUE #19: Add async variants of `XPCClient.send/sendMessage(...)` functions

jakaplan:
Fixes #19

Adds new tests for the async variants. Updates documentation to match.

### DIFF
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -225,7 +225,7 @@ public class XPCClient {
     ///   - route: The server route which will handle this.
     @available(macOS 10.15.0, *)
     public func send(route: XPCRouteWithoutMessageWithoutReply) async throws {
-        try await withCheckedThrowingContinuation { continuation in
+        try await withUnsafeThrowingContinuation { continuation in
             send(route: route) { response in
                 continuation.resume(with: response)
             }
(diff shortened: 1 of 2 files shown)


jakaplan:
@amomchilov for your review

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -191,6 +219,19 @@ public class XPCClient {
         }
     }
     
+    /// Sends with no message and does not receive a reply.
+    ///
+    /// - Parameters:
+    ///   - route: The server route which will handle this.
+    @available(macOS 10.15.0, *)
+    public func send(route: XPCRouteWithoutMessageWithoutReply) async throws {
+        try await withCheckedThrowingContinuation { continuation in
```
amomchilov, line 228: Why `checked`? This looks safe for `withUnsafeThrowingContinuation`

Tests/SecureXPCTests/Client & Server/Round-trip Integration Test.swift
```diff
@@ -20,7 +20,7 @@ class RoundTripIntegrationTest: XCTestCase {
         anonymousServer.start()
     }
 
-    func testSendWithMessageWithReply() throws {
+    func testSendWithMessageWithReply_Sync() throws {
```
amomchilov, line 23: I think it's reasonable to just completely replace these closure-based tests.

In my own work, I've pretty much completely abandoned the XCTest continuation APIs in favour of async/await, even for old projects that don't use it in the production code (I just add wrappers in my test bundles, which makes test writing easier and more correct).

amomchilov (approved):

Tests/SecureXPCTests/Client & Server/Round-trip Integration Test.swift
jakaplan: I'm going to leave them for now as the behavior differs for two of them depending on whether they have a closure provided or not, but point taken.

PR #63, opened 2022-01-20
Makes creating anonymous servers part of the public API

jakaplan:
Almost all of these changes are documentation related aside from changing the access levels from `internal` to `public`

### DIFF
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -13,8 +13,9 @@ import Foundation
 /// There are two different types of services you can communicate with using this client: XPC Services and XPC Mach services. If you are uncertain which
 /// type of service you're using, it's likely an XPC Service.
 ///
-/// **XPC Services**
+/// Clients can also be created from an ``XPCServerEndpoint`` which is the only way to create a client for an anonymous server.
 ///
+/// #### XPC Services
 /// These are helper tools which ship as part of your app and only your app can communicate with.
 ///
 /// The name of the service must be specified when retrieving a client to talk to your XPC Service; this is always the bundle identifier for the service:
@@ -25,8 +26,7 @@ import Foundation
 /// The service itself must create and configure an ``XPCServer`` by calling ``XPCServer/forThisXPCService()`` in order for this client to be able to
 /// communicate with it.
 ///
-/// **XPC Mach services**
-///
+/// #### XPC Mach services
 /// Launch Agents, Launch Daemons, and helper tools installed with
 /// [  `SMJobBless`](https://developer.apple.com/documentation/servicemanagement/1431078-smjobbless) can optionally communicate
 /// over XPC by using Mach services.
@@ -37,6 +37,14 @@ import Foundation
 /// ```
 /// The tool itself must retrieve and configure an ``XPCServer`` by calling ``XPCServer/forThisMachService(named:clientRequirements:)`` or
 /// ``XPCServer/forThisBlessedHelperTool()`` in order for this client to be able to communicate with it.
+///
+/// #### Endpoints
+/// Clients can be created from server endpoints:
+/// ```swift
+/// let server = XPCServer.makeAnonymous()
+/// let client = XPCClient.forEndpoint(server.endpoint)
+/// ```
+/// Server endpoints can also be sent across XPC connections.
 ///
 /// ### Calling Routes
 /// Once a client has been retrieved, calling a route is as simple as invoking `send` with a route:
@@ -162,6 +170,9 @@ public class XPCClient {
         XPCMachClient(machServiceName: machServiceName)
     }
 
+    /// Provides a client to communicate with the server corresponding to the provided endpoint.
+    ///
+    /// A server's endpoint is accesible via ``NonBlockingServer/endpoint``. The endpoint can be sent across an XPC connection.
 	public static func forEndpoint(_ endpoint: XPCServerEndpoint) -> XPCClient {
         let connection = xpc_connection_create_from_endpoint(endpoint.endpoint)
 
--- a/Sources/SecureXPC/Server/XPCServer.swift
+++ b/Sources/SecureXPC/Server/XPCServer.swift
@@ -13,6 +13,8 @@ import Foundation
 /// There are two different types of services you can retrieve a server for: XPC Services and XPC Mach services. If you're uncertain which type of service you're
 /// using, it's likely an XPC Service.
 ///
+/// Anonymous servers can also be created which do not correspond to a service.
+///
 /// #### XPC Services
 /// These are helper tools which ship as part of your app and only your app can communicate with.
 ///
@@ -22,7 +24,6 @@ import Foundation
 /// ```
 ///
 /// #### XPC Mach services
-///
 /// Launch Agents, Launch Daemons, and helper tools installed with
 /// [`SMJobBless`](https://developer.apple.com/documentation/servicemanagement/1431078-smjobbless) can optionally communicate
 /// over XPC by using Mach services.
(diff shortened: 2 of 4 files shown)


jakaplan:
@amomchilov In terms of reviewing, let me know if you think there's anything remaining that needs to happen before making anonymous servers public. Your feedback on the documentation is also welcome, but completely fine if you don't review that.

Sources/SecureXPC/XPCServerEndpoint.swift
```diff
@@ -1,12 +1,18 @@
...
 //
 
 import Foundation
 
+/// An endpoint is used to create clients which can communicate with their associated server.
+///
+/// Endpoints are retrieved from a server's ``NonBlockingServer/endpoint`` property. They can be used in the same process or sent across an existing XPC
+/// connection.
+///
+/// > Warning: While ``XPCServerEndpoint`` conforms to `Codable` it can only be encoded and decoded by the `SecureXPC` framework.
```
amomchilov, line 15: The Swift doc format has a style exactly for this:

```suggestion
/// - Warning:
```

Perhaps `Important` might be more suitable, IDK. You can see them all here: https://nshipster.com/swift-documentation/#additional-fields

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -13,8 +13,9 @@ import Foundation
 /// There are two different types of services you can communicate with using this client: XPC Services and XPC Mach services. If you are uncertain which
 /// type of service you're using, it's likely an XPC Service.
 ///
-/// **XPC Services**
+/// Clients can also be created from an ``XPCServerEndpoint`` which is the only way to create a client for an anonymous server.
```
amomchilov, line 16:
```suggestion
/// Clients can also be created from an ``XPCServerEndpoint``, which is the only way to create a client for an anonymous server.
```

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -13,6 +13,8 @@ import Foundation
 /// There are two different types of services you can retrieve a server for: XPC Services and XPC Mach services. If you're uncertain which type of service you're
 /// using, it's likely an XPC Service.
 ///
+/// Anonymous servers can also be created which do not correspond to a service.
```
amomchilov, line 16: This is a bit jargony IMO. I'm not sure "a server without a service" would make sense unless people were already in-the-know.

Perhaps:

```suggestion
/// Anonymous servers can also be created, which do not correspond to a Mach or XPC service. These can be useful for testing, or to aid in development by removing the need to debug across multiple processes.
```

amomchilov: NVM, you cover use cases below, but you get the idea

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -312,7 +322,9 @@ public class XPCServer {
 public protocol NonBlockingServer {
     /// Begins processing requests received by this XPC server.
     func start()
-    /// Use an endpoint to create connections to this server
+    /// Use an endpoint to create connections to this server.
```
amomchilov, line 325: This verb-phrase doesn't really cover what this property does itself (it talks about what you can do down the line). How about splitting that up, such as:

```suggestion
    /// Export an endpoint for this server, which can later be used to create new connections to this server.
```

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -341,12 +353,56 @@ extension XPCServer {
...
+    /// Creates a new anonymous server that accepts requests from the same process it's running in.
+    ///
+    /// Only a client created from an anonymous server's endpoint can communicate with that server. Retrieve the ``NonBlockingServer/endpoint`` and
+    /// then create a client with it:
+    /// ```swift
+    /// let server = XPCServer.makeAnonymous()
+    /// let client = XPCClient.fromEndpoint(server.endpoint)
+    /// ```
+    ///
+    /// > Important: No requests will be processed until ``NonBlockingServer/start()`` or ``startAndBlock()`` is called.
```
amomchilov, line 365:
```
    /// - Important:
```

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -341,12 +353,56 @@ extension XPCServer {
...
+    ///
+    /// ## Usage
+    /// If you need this server to be communicated with by clients running in a different process, use ``makeAnonymous(clientRequirements:)`` instead.
```
amomchilov, line 368: This isn't really "usage" information, it's kind of the opposite. Perhaps this should be an `Important`, `Attention`, `Note` or `Remark`?

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -341,12 +353,56 @@ extension XPCServer {
...
+    public static func makeAnonymous() -> XPCServer & NonBlockingServer {
         XPCAnonymousServer(messageAcceptor: SameProcessMessageAcceptor())
     }
 
-    internal static func makeAnonymous(clientRequirements: [SecRequirement]) -> XPCServer & NonBlockingServer {
+    /// Creates a new anonymous server that accepts requests from clients which meet the security requirements.
+    ///
+    /// Only a client created from an anonymous server's endpoint can communicate with that server. Retrieve the ``NonBlockingServer/endpoint`` and
```
amomchilov, line 375: This is a bit of a sentence fragment, and doesn't fit in:


> "Retrieve the ``NonBlockingServer/endpoint`` and send it across an existing XPC connection."

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -341,12 +353,56 @@ extension XPCServer {
...
+    /// ```swift
+    /// let reqString = "identifier \"com.example.AuthorizedClient\" and certificate leaf[subject.OU] = \"4L0ZG128MM\""
+    /// var requirement: SecRequirement?
+    /// if SecRequirementCreateWithString(reqString as CFString,
+    ///                                   SecCSFlags(),
+    ///                                   &requirement) == errSecSuccess,
+    ///   let requirement = requirement {
+    ///     let server = XPCServer.makeAnonymous(clientRequirements: [requirement])
+    ///
+    ///    <# configure and start server #>
```
amomchilov, line 389: Indentation:

```suggestion
    ///     let requirement = requirement {
    ///     let server = XPCServer.makeAnonymous(clientRequirements: [requirement])
    ///
    ///     <# configure and start server #>
```

amomchilov (approved):
Some, comments, though it's mainly subjective grammar stuff

Sources/SecureXPC/XPCServerEndpoint.swift
jakaplan: This is the [DocC format](https://developer.apple.com/documentation/xcode/formatting-your-documentation-content):
image

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: I believe a comma in this case wouldn't be appropriate as the clause which follows is essential information. See https://www.grammarly.com/blog/comma-before-which/.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Good call, will incorporate the first portion of this.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Indeed, incorporated your feedback.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Same note about this being the DocC format.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Hmm, yeah I was unsure of how to refer to this. I'll use a `Note`.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Yeah it does read a bit awkwardly, I've changed it to:
> Do this by retrieving the server’s `NonBlockingServer/endpoint` and then creating a client with it:

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Good catch, fixed.

Sources/SecureXPC/XPCServerEndpoint.swift
amomchilov: TIL. Nice!

PR #64, opened 2022-01-21
Allows for async functions/closures to be registered as routes

jakaplan:
Adds tests to match


jakaplan:
@amomchilov For your review. I dislike how much code duplication this introduces, but I couldn't figure out a way to reduce it due to the cascading nature of `async`.

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -128,6 +133,13 @@ public class XPCServer {
             connections.compactMap{ $0.connection }.forEach{ xpc_connection_set_target_queue($0, newValue) }
         }
     }
+    // MARK: Route registration
+        
+    private func checkUniqueRoute(route: XPCRoute) throws {
+        if self.routes.keys.contains(route) {
```
amomchilov, line 139: IDK how much it matters, but these two operations (checking it's not already there, and setting the new value) could be done in a single hash look-up with https://developer.apple.com/documentation/swift/dictionary/3127179-updatevalue

amomchilov (commented):
I'll look at this in more depth over the weekend.

Until then, could you help me understand when a callback should be `async`?

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Eh yeah I don't think any efficiency gains matter here, I'd rather consolidate the error generation into just one place.

jakaplan:
> Until then, could you help me understand when a callback should be `async`?

When the API user wants to create a callback which uses `async` functions within it, which will be an increasing percentage of functions over time. Without this the caller can't have any route with a reply (or one they want an error to be propagated back from) use `async` functions because wrapping the code in a `Task` won't allow for the result (or thrown error) to be returned.

jakaplan:
A realistic example could be an XPC service with a route provided a URL to a PDF. The service needs to download and then parse the PDF to extract some data. PDFs are a notoriously exploit vector, so breaking out this component into an XPC service would be prudent. Using `async` functions to perform the download would be easiest. And there very well could be `async` functions involved in the data extraction (ex. running ML inference over it to extract text embedded in images).

amomchilov:
> When the API user wants to create a callback which uses `async` functions within it, which will be an increasing percentage of functions over time. Without this the caller can't have any route with a reply (or one they want an error to be propagated back from) use `async` functions because wrapping the code in a `Task` won't allow for the return value (or thrown error) to be returned.

Ah yes, that makes perfect sense, thanks for the explanation

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -137,11 +149,21 @@ public class XPCServer {
     /// - Throws: If this route has already been registered.
     public func registerRoute(_ route: XPCRouteWithoutMessageWithoutReply,
                               handler: @escaping () throws -> Void) throws {
```
amomchilov, line 151: In retrospect, I don't think throwing here was a good idea (introduced in #28).

These errors can't reasonably recovered:

1. "Do nothing" => the route doesn't act the way you expect
2. "Name it something different" => makes it useless to the callers that expect a given name

Fatal erroring here might be more appropriate, so that route-registration doesn't need a `try` (removing `try` for brevity is a non-goal for me, but in some cases an unexpected `try` raises a suspicion error was thrown that can't really be caught)

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -152,11 +174,21 @@ public class XPCServer {
...
-        if self.routes.keys.contains(route.route) {
-            throw XPCError.routeAlreadyRegistered(route.route.pathComponents)
-        }
-        
-        self.routes[route.route] = ConstrainedXPCHandlerWithMessageWithoutReply(handler: handler)
+        try checkUniqueRoute(route: route.route)
+        self.routes[route.route] = ConstrainedXPCHandlerWithMessageWithoutReplySync(handler: handler)
+    }
+    
+    /// Registers a route that has a message and can't receive a reply.
```
amomchilov, line 181: I would use more context-specific terminology:

> /// Registers a route that sends message and can't receive a reply.

This is similar to how below you say "expects a reply" rather than "has a reply"

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -219,15 +271,10 @@ public class XPCServer {
     
     private func handleEvent(connection: xpc_connection_t, event: xpc_object_t) {
         if xpc_get_type(event) == XPC_TYPE_DICTIONARY {
```
amomchilov, line 273: Another case where my guard clause instinct tingles:


```swift
    private func handleEvent(connection: xpc_connection_t, event: xpc_object_t) {
        // Only dictionary types (and not anything which will be interpreted as an error) should be checked for
        // message acceptance as errors generated by libxpc can fail to meet the acceptor's criteria
        guard xpc_get_type(event) == XPC_TYPE_DICTIONARY else {
            self.errorHandler?(XPCError.fromXPCObject(event))
            return
        }

        guard self.messageAcceptor.acceptMessage(connection: connection, message: event) else {
            self.errorHandler?(.insecure)
            return
        }

        handleMessage(connection: connection, message: event)
    }
```

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -219,15 +271,10 @@ public class XPCServer {
...
             if self.messageAcceptor.acceptMessage(connection: connection, message: event) {
-                var reply = xpc_dictionary_create_reply(event)
-                do {
-                    try handleMessage(connection: connection, message: event, reply: &reply)
-                } catch {
-                    let error = XPCError.asXPCError(error: error, expectingOtherError: true)
-                    self.errorHandler?(error)
-                    self.replyWithErrorIfPossible(error, connection: connection, reply: &reply)
-                }
+                handleMessage(connection: connection, message: event)
```
amomchilov, line 277: This is nice 🧹

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -236,19 +283,64 @@ public class XPCServer {
...
+                    // If a dictionary reply exists, then the message expects a reply to be sent back
+                    if let reply = reply {
+                        xpc_connection_send_message(connection, reply)
+                    }
+                } catch {
+                    self.handleError(error, expectingOtherError: true, connection: connection, reply: &reply)
+                }
+            }
+        } else {
+            fatalError("Server handler for route \(request.route.pathComponents) cannot be run: \(handler)")
```
amomchilov, line 331: Could you clarify this case?

By my reading, it happens if you have a `XPCHandlerAsync` registered, but don't pass the `#available(macOS 10.15.0, *)` check?

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -236,19 +283,64 @@ public class XPCServer {
...
+                
+                // If a dictionary reply exists, then the message expects a reply to be sent back
+                if let reply = reply {
+                    xpc_connection_send_message(connection, reply)
+                }
+            } catch {
+                self.handleError(error, expectingOtherError: true, connection: connection, reply: &reply)
+            }
+            
+        } else if #available(macOS 10.15.0, *), let handler = handler as? XPCHandlerAsync {
```
amomchilov, line 316: Oh cool, I didn't know these compile-time and run-time checks could be combined into one statement like this. Neat!

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -236,19 +283,64 @@ public class XPCServer {
...
+            Task {
+                var reply = xpc_dictionary_create_reply(message)
+                do {
+                    try await handler.handle(request: request, reply: &reply)
+                    
+                    // If a dictionary reply exists, then the message expects a reply to be sent back
+                    if let reply = reply {
+                        xpc_connection_send_message(connection, reply)
+                    }
```
amomchilov, line 325: Should this be moved out of this `do` block and after the catch below?

amomchilov: NVM, I think my bias towards minimal `do` blocks comes from Java/Ruby/etc. where any one of the lines can be the source of the thrown exception, and it's unclear which ones you intend to catch.

I have a new-found appreciation for `try`. It makes this so clear.

amomchilov (approved):

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Hmm, this was my attempt to be specific. It's not that it doesn't expect a reply, it very much cannot have one - it's quite literally impossible with the way the code is written.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Yeah. It shouldn't be possible, but seemed prudent to explicitly fail if it was encountered. I can improve the error message?

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Yeah, it's very convenient!

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Hmm maybe it's just my relative lack of usage of `guard`, but I find this a bit harder to read.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: > Should this be moved out of this `do` block and after the catch below?

I know you said never mind, but FWIW it needs to be in the `do` block as it should only run if the handler succeeds. Otherwise both this logic and the logic in the `catch` block would run resulting in two sends to the same `reply` instance which would result in XPC API Misuse.

Sources/SecureXPC/Server/XPCServer.swift
amomchilov: Just explain why it doesn't makes sense, along the lines of "The route wasn't an sync route, even that's the only kind that should be possible for this system"

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: Let me think about this. If I make this change, will do so as part of a separate PR.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: On second thought, I do prefer this. Have updated the comment to make a bit more sense with this control flow.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: 🧼

PR #65, opened 2022-01-21
`XPCRoute` builder style creation

jakaplan:
Implementation based on discussion #32


jakaplan:
@amomchilov would appreciate you reviewing this. I'm pretty sure this is what you had in mind from discussion #32. I figured this would be nice to include in the next major versioned release.

amomchilov:
Ahhh, you should have told me you wanted to go in this direction, I would have opened a new PR for the POC that I had linked in that discussion

Sources/SecureXPC/Routes.swift
```diff
@@ -7,8 +7,57 @@
 
 import Foundation
 
-/// Consistent framework internal implementation of routes that can be sent over XPC (because its Codable) and used as a dictionary key (because its Hashable).
-struct XPCRoute: Codable, Hashable {
+/// The entry point for creating routes needed for client to server communication.
+///
+/// Client to server communication is performed using routes. A route is a sequence of `String`s and if applicable also the message and reply types.
+///
+/// In practice a route is similar to a function signature or a server path with type safety, although it is not precisely either. Message and reply types must be
```
amomchilov, line 14: > although it is not precisely either

Reads awkwardly, as "(it is not precisely) either" rather than "it is not (precisely either)"

Sources/SecureXPC/Routes.swift
```diff
@@ -7,8 +7,57 @@
...
+///
+/// #### Registering Routes
+/// Once a route has been created, it needs to be registered with an ``XPCServer`` in order for it to be called by an ``XPCClient``.
+///
+/// #### Updating Routes
+/// If you are using XPC Mach services, take care when updating existing routes because over time you may end up with an
+/// older version of your server installed on a computer with a newer client. You may want to version your routes by
+/// prefixing them with a number:
+/// ```swift
+/// XPCRoute.named("1", "reset")
```
amomchilov, line 56: Brain-storming: once we have more real world experience maintaining these, perhaps we should build versioning/fallbacks/etc. as a first class feature of the framework

amomchilov (approved):
Love it

Sources/SecureXPC/Routes.swift
jakaplan: Thanks for the feedback, will just remove this whole portion of the sentence

Sources/SecureXPC/Routes.swift
jakaplan: The thought I've had in this space is to create built-in support for a client to query if a server has a certain route registered.

jakaplan:
> Ahhh, you should have told me you wanted to go in this direction, I would have opened a new PR for the POC that I had linked in that discussion

Sorry about that, I thought my expression of interest/support was clear and that you were just busy. Will be more explicit in the future.

Sources/SecureXPC/Routes.swift
amomchilov: I don't think that would be sufficient, because the name/msg-type/reply-type could all be the same, but they might still be incompatible due to changes in the msg/reply types.

Now that I think abotu it, that's more of a "Codable doesn't have versioning facilities" problem, not necessarily specific to SecureXPC.

amomchilov:
Ah, sorry for the confusing. I ccouldn't tell "what's a vaguely a good idea for some point in the future" vs "imminently useful"

Sources/SecureXPC/Routes.swift
jakaplan: Yeah checking for registered routes wouldn't be something automatic, but combined with proper versioning could work well.  Quite possibly there are better ways to achieve this. 

In terms of `Codable` the issue is that actually the types don't need to match across client and server for it to be valid. As long as encoding/decoding is possible, it's fine. And in some ways this is a good thing for example if the type is renamed, but not otherwise changed. I don't know there's a way to validate an encoding/decoding is possible without actually trying it. Haven't given this much thought though.

Precise versioning control is why things like gRPC exist. (I actually briefly contemplated using gRPC instead of `Codable`.)

PR #68, opened 2022-01-24
Makes repeated route registration result in `fatalError` instead of an `Error`

jakaplan:
In doing so, slightly consolidates the internal implementation of route registration.

### DIFF
--- a/Sources/SecureXPC/Server/XPCServer.swift
+++ b/Sources/SecureXPC/Server/XPCServer.swift
@@ -83,15 +83,15 @@ import Foundation
 /// - ``makeAnonymous()``
 /// - ``makeAnonymous(clientRequirements:)``
 /// ### Registering Routes
-/// - ``registerRoute(_:handler:)-82935``
-/// - ``registerRoute(_:handler:)-7yvyr``
-/// - ``registerRoute(_:handler:)-3ohmq``
-/// - ``registerRoute(_:handler:)-4jjs6``
+/// - ``registerRoute(_:handler:)-4ttqe``
+/// - ``registerRoute(_:handler:)-9a0x9``
+/// - ``registerRoute(_:handler:)-4fxv0``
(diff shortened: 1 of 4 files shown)


jakaplan:
Follow up from feedback for #64 

@amomchilov for your review

jakaplan:
Also @amomchilov do you know of any (non-hacky) way to test `fatalError`? It'd be nice to write a unit test or two that ensured that `fatalError` is in fact occurring in the case of duplicative route registration.

amomchilov:
@jakaplan It's possible, but it's currently one of my biggest pain points in Swift.

Check out https://github.com/mattgallagher/CwlPreconditionTesting. It has an accompanying blog post series: https://www.cocoawithlove.com/blog/2016/02/02/partial-functions-part-two-catching-precondition-failures.html

I've never used `CwlPreconditionTesting` directly, only through the [`throwAssertion()` matcher in Nimble](https://github.com/Quick/Nimble/#swift-assertions).

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -144,112 +144,117 @@ public class XPCServer {
...
     // MARK: Route registration
         
-    private func checkUniqueRoute(route: XPCRoute) throws {
-        if self.routes.keys.contains(route) {
-            throw XPCError.routeAlreadyRegistered(route.pathComponents)
+    /// Internal function that actually registers the route and enforces that a route is only ever registered once.
+    ///
+    /// All of the public functions exist to satisfy type constraints.
+    private func registerRoute(_ route: XPCRoute, handler: XPCHandler) {
+        if let _ = self.routes.updateValue(handler, forKey: route) {
```
amomchilov, line 154: I think this reads more clearly:

```suggestion
    if self.routes.updateValue(handler, forKey: route) != nil {
```

amomchilov (approved):
Looks reasonable to me.

PR #69, opened 2022-01-24
Misc improvements throughout the code base

jakaplan:
I realized they're actually called "XPC services" not "XPC Services", so have updated that throughout the comments (both public and internal)

The rest of the the changes are all misc code changes that I think improve things for the better. There's one small public facing change to `XPCError.misconfiguredXPCService` which now has a String to provide a descriptive error message.

### DIFF
--- a/Sources/SecureXPC/Server/XPCMachServer.swift
+++ b/Sources/SecureXPC/Server/XPCMachServer.swift
@@ -190,23 +190,21 @@ extension XPCMachServer {
     /// - Returns: The property list as data.
     private static func readEmbeddedPropertyList(sectionName: String) throws -> Data {
         // By passing in nil, this returns a handle for the dynamic shared object (shared library) for this helper tool
-        if let handle = dlopen(nil, RTLD_LAZY) {
-            defer { dlclose(handle) }
-
-            if let mhExecutePointer = dlsym(handle, MH_EXECUTE_SYM) {
-                let mhExecuteBoundPointer = mhExecutePointer.assumingMemoryBound(to: mach_header_64.self)
-
-                var size = UInt(0)
-                if let section = getsectiondata(mhExecuteBoundPointer, "__TEXT", sectionName, &size) {
-                    return Data(bytes: section, count: Int(size))
-                } else { // No section found with the name corresponding to the property list
-                    throw XPCError.misconfiguredBlessedHelperTool("Missing property list section \(sectionName)")
-                }
-            } else { // Can't get pointer to MH_EXECUTE_SYM
-                throw XPCError.misconfiguredBlessedHelperTool("Could not read property list (nil symbol pointer)")
-            }
-        } else { // Can't open handle
+        guard let handle = dlopen(nil, RTLD_LAZY) else {
             throw XPCError.misconfiguredBlessedHelperTool("Could not read property list (handle not openable)")
         }
+        defer { dlclose(handle) }
+        
+        guard let mhExecutePointer = dlsym(handle, MH_EXECUTE_SYM) else {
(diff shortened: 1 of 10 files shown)


jakaplan:
@amomchilov for your review

Sources/SecureXPC/Server/XPCMachServer.swift
```diff
@@ -190,23 +190,21 @@ extension XPCMachServer {
...
-                if let section = getsectiondata(mhExecuteBoundPointer, "__TEXT", sectionName, &size) {
-                    return Data(bytes: section, count: Int(size))
-                } else { // No section found with the name corresponding to the property list
-                    throw XPCError.misconfiguredBlessedHelperTool("Missing property list section \(sectionName)")
-                }
-            } else { // Can't get pointer to MH_EXECUTE_SYM
-                throw XPCError.misconfiguredBlessedHelperTool("Could not read property list (nil symbol pointer)")
-            }
-        } else { // Can't open handle
+        guard let handle = dlopen(nil, RTLD_LAZY) else {
```
amomchilov, line 193: 😍 

Is `guard` starting to read a bit more naturally for you, yet?

Tests/SecureXPCTests/Client & Server/Server Termination Integration Test.swift
```diff
@@ -19,52 +19,45 @@ class ServerTerminationIntegrationTest: XCTestCase {
...
-        // Make two more calls after having shutdown the server:
-        // - The second call should fail upon trying to send the message, connection is interrupted
-        // - The third call (and any subsequent ones) should fail indicating no new connections can ever be established
         let interruptedExpectation = self.expectation(description: "Second message results in an interrupted error")
-        let cannotBeReestablishedExpectation = self.expectation(description: "Third message can't be sent")
-        client.sendMessage("2nd message", route: echoRoute) { response in
-            switch response {
-                case .failure(.connectionInterrupted):
+        do {
+            try await client.send(route: route)
```
amomchilov, line 38: Ah, looks like [`XCTAssertThrowsError(_:_:file:line:_:)`](https://developer.apple.com/documentation/xctest/1500795-xctassertthrowserror) can't take an `async` expression, eh?

amomchilov (approved):

Sources/SecureXPC/Server/XPCMachServer.swift
jakaplan: Yeah I'm increasingly coming around to it. I'm used to a much more "branching" style of code, which while objectively is probably harder to reason about - is what I've implicitly trained myself to well understand.

Tests/SecureXPCTests/Client & Server/Server Termination Integration Test.swift
jakaplan: Yeah. As I just learned, making `async` versions of everything is rather a pain. I figure it's going to be a multiyear journey until everything that ought to be or support Swift concurrency does. Certainly complicated by the fact that Objective-C doesn't have a direct equivalent and it's going to be important for large established code bases for many years to come.

amomchilov: Looks like [Quick](https://github.com/Quick/Quick) doesn't support it [yet](https://github.com/Quick/Quick/issues/1084), either.

I have larges for Quick to become the standard testing framework in the Swift ecosystem. It happens to use XCTest under the hood, but pretty much entirely decouples you from it, to isolate you from the madness (e.g. `XCAssertEqual` makes no indication as to which param should be "expected" vs "actual", and a ton of codebases (including mine) use it wrong/inconsistently T_T)

PR #71, opened 2022-01-25
Always sets connection to nil when interrupted

### LINKED ISSUE #70: `XPC_ERROR_CONNECTION_INTERRUPTED` doesn't appear to actually behave as documented
After releasing 0.3.0 of SecureXPC I started updating [SwiftAuthorizationSample](https://github.com/trilemma-dev/SwiftAuthorizationSample) and have run into an unexpected issue:
1. Over XPC the app asks the helper tool to uninstall itself
2. As part of the helper tool uninstalling itself, the helper tool terminates
3. This results in the client receiving `XPC_ERROR_CONNECTION_INTERRUPTED`.
  i. This is expected, Apple's documentation for interrupted says: "Will be delivered to the connection’s event handler if the remote service exited."
4. Installer the helper tool again (under the hood via `SMJobBless`).
5. The next `send` using the client results in a `XPC_ERROR_CONNECTION_INVALID`
  i. This is **unexpected**, Apple's documentation for interrupted also says: "The connection is still live even in this case, and resending a message will cause the service to be launched on-demand."
 ii. I can confirm the service is **not** launched on-demand.
6. The subsequent `send` using the client results in the service being launched.
 i. This is not because of anything XPC is doing automatically, it's because when we received `XPC_ERROR_CONNECTION_INVALID` in step 4 we perform `self.connection = nil` which means we establish a new connection on the next `send`.

jakaplan:
Fixes #70

Updates detailed comment to explain why this is

### DIFF
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -435,15 +435,13 @@ public class XPCClient {
             //   Will be delivered to the connection’s event handler if the remote service exited. The connection is
             //   still live even in this case, and resending a message will cause the service to be launched on-demand.
             //
-            // While Apple's documentation is technically correct, it's misleading in the case of an anonymous
-            // connection where there is no service. Because there is no service, there is nothing to be relaunched
-            // on-demand. The connection might technically still be alive, but resending a message will *not* work.
+            // From observed behavior Apple's documentation is *not* correct. After the connection is interrupted the
+            // subsequent call will result in XPC_ERROR_CONNECTION_INVALID and the service will not be relaunched.
             //
-            // By setting the connection to `nil` when there is no service (indicated by no service name), anonymous
-            // clients can throw a useful specific error when `createConnection()` is called.
-            if self.serviceName == nil {
-                self.connection = nil
-            }
+            // Additionally, in the case of an anonymous connection there is no service. Because there is no service,
+            // there is nothing to be relaunched on-demand. The connection might technically still be alive, but
+            // resending a message will *not* work.
+            self.connection = nil
         }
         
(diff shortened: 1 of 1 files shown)


jakaplan:
I tested this branch against the scenario described in #70 and confirmed it now works as expected, the `send` in step 4 now succeeds.

jakaplan:
@amomchilov for your review. It'd be beneficial reading through #70 first for context.

jakaplan:
@amomchilov friendly ping if you have a few minutes to take a look

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -435,15 +435,13 @@ public class XPCClient {
             //   Will be delivered to the connection’s event handler if the remote service exited. The connection is
             //   still live even in this case, and resending a message will cause the service to be launched on-demand.
             //
-            // While Apple's documentation is technically correct, it's misleading in the case of an anonymous
-            // connection where there is no service. Because there is no service, there is nothing to be relaunched
-            // on-demand. The connection might technically still be alive, but resending a message will *not* work.
+            // From observed behavior Apple's documentation is *not* correct. After the connection is interrupted the
+            // subsequent call will result in XPC_ERROR_CONNECTION_INVALID and the service will not be relaunched.
```
amomchilov, line 439: I think it's worth dropping a link in here to `https://github.com/trilemma-dev/SecureXPC/issues/70`, which can elaborate further.

amomchilov (approved):
This looks fine, I just don't know if it's the appropriate solution to the problem (which I don't fully understand yet), but we can discuss that in #70

Sources/SecureXPC/Client/XPCClient.swift
jakaplan: Will do

PR #73, opened 2022-02-23
Adds `XPCRequestContext` which exposes client info to the server

jakaplan:
The core addition here is the introduction of the entirely optional `XPCRequestContext` which has static properties that are accessible within the execution context of a closure called by `XPCServer`.

The motivation behind this PR is that there are times where I need to differ the behavior of the server based on the caller. Since I trust all of my clients I could have them explicitly provide this information, but it seemed more robust to have this be determined server side.

The design is a bit "magic" so I'm torn on it. I considered three different approaches and ultimately settled on this one. The other two considered were:
1. Closures registered with the server *must* have `XPCRequestContext` as a parameter.
2. Closures registered with the server *may* have `XPCRequestContext` as a parameter.

Approach 1 felt far too heavy handed as I don't think the average user of SecureXPC will need to make use of `XPCRequestContext`. Approach 2 would require doubling the number of handler registration functions from 8 to 16 along with all of the associated boiler plate. Additionally it'd make `XPCRequestContext` more prominent in the API and that's undesirable as again I think it'll be needed in a small minority of cases.

### DIFF
--- a/Sources/SecureXPC/XPCRequestContext.swift
+++ b/Sources/SecureXPC/XPCRequestContext.swift
@@ -14,8 +14,8 @@ import Foundation
 /// ## Topics
 /// ### Client Information
 /// - ``clientCode``
-/// - ``effectiveUserId``
-/// - ``effectiveGroupId``
+/// - ``effectiveUserID``
+/// - ``effectiveGroupID``
 public class XPCRequestContext {
     private let connection: xpc_connection_t
     private let message: xpc_object_t
@@ -26,7 +26,7 @@ public class XPCRequestContext {
     }
     
     // MARK: task local
-        
+    
     @available(macOS 10.15.0, *)
     @TaskLocal
     private static var currentForTask: XPCRequestContext?
@@ -77,12 +77,12 @@ public class XPCRequestContext {
     // readily result in security vulnerabilities
     
     /// The effective user id of the client process.
-    public static var effectiveUserId: uid_t {
+    public static var effectiveUserID: uid_t {
         xpc_connection_get_euid(current.connection)
     }
     
     /// The effective group id of the client process.
-    public static var effectiveGroupId: gid_t {
+    public static var effectiveGroupID: gid_t {
         xpc_connection_get_egid(current.connection)
     }
     
--- a/Sources/SecureXPC/XPCRequestContext.swift
+++ b/Sources/SecureXPC/XPCRequestContext.swift
@@ -1,0 +1,93 @@
+//
+//  XPCRequestContext.swift
+//  SecureXPC
+//
+//  Created by Josh Kaplan on 2022-02-22.
+//
+
+import Foundation
+
+/// Provides information about the current request being handled by an ``XPCServer``.
+///
+/// Accessing the properties exposed by this class is a programming error unless called from inside of a closure registered with and called by an `XPCServer`.
+///
+/// ## Topics
+/// ### Client Information
+/// - ``clientCode``
+/// - ``effectiveUserId``
+/// - ``effectiveGroupId``
+public class XPCRequestContext {
+    private let connection: xpc_connection_t
+    private let message: xpc_object_t
+    
+    private init(connection: xpc_connection_t, message: xpc_object_t) {
+        self.connection = connection
+        self.message = message
+    }
+    
+    // MARK: task local
+        
+    @available(macOS 10.15.0, *)
+    @TaskLocal
+    private static var currentForTask: XPCRequestContext?
+    
+    @available(macOS 10.15.0, *)
+    @discardableResult internal static func setForTask<Success, Failure>(
+        connection: xpc_connection_t,
+        message: xpc_object_t,
+        operation: () throws -> Task<Success, Failure>
+    ) rethrows -> Task<Success, Failure> {
+        try $currentForTask.withValue(XPCRequestContext(connection: connection, message: message)) {
+            try operation()
+        }
+    }
+    
+    // MARK: thread local
+    
+    private static let contextKey = UUID()
+    
+    @discardableResult internal static func setForCurrentThread<R>(
+        connection: xpc_connection_t,
+        message: xpc_object_t,
+        operation: () throws -> R
+    ) rethrows -> R {
+        Thread.current.threadDictionary[contextKey] = XPCRequestContext(connection: connection, message: message)
+        let result = try operation()
+        Thread.current.threadDictionary.removeObject(forKey: contextKey)
+        
+        return result
+    }
+    
+    // MARK: current context
+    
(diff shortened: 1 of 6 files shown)


jakaplan:
@amomchilov would appreciate your feedback on this, both at a code level as well as the overall approach. I think this is reasonable design, but I'm not entirely convinced it's the best one.

Sources/SecureXPC/XPCRequestContext.swift
```diff
@@ -0,0 +1,93 @@
...
+        connection: xpc_connection_t,
+        message: xpc_object_t,
+        operation: () throws -> R
+    ) rethrows -> R {
+        Thread.current.threadDictionary[contextKey] = XPCRequestContext(connection: connection, message: message)
+        let result = try operation()
+        Thread.current.threadDictionary.removeObject(forKey: contextKey)
+        
+        return result
+    }
```
amomchilov, line 59: I would extract this concern into your own property wrapper, which automatically picks between task local or thread local storage. IDK what I would call it though, naming is hard.

Sources/SecureXPC/XPCRequestContext.swift
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+    
+    // MARK: public functions
+    
+    // It's intentional that the process id (PID) of the client process is not exposed since misuse of this can
+    // readily result in security vulnerabilities
+    
+    /// The effective user id of the client process.
+    public static var effectiveUserId: uid_t {
```
amomchilov, line 80: Apple/Swift's convention is make the "d" have the same capitalization as the "i". E.g. `var id` or `var effectiveUserID`. Likewise below

Tests/SecureXPCTests/Client & Server/XPCRequestContext Tests.swift
```diff
@@ -0,0 +1,131 @@
...
+    func testGetClientCode_sync() throws {
+        let route = XPCRoute.named("does", "nothing")
+        let server = XPCServer.makeAnonymous()
+        let client = XPCClient.forEndpoint(server.endpoint)
+        
+        let clientCodeNotNil = self.expectation(description: "Client code should not be nil in this circumstance")
+        let xctestIdentifier = self.expectation(description: "Client code's identifier should be com.apple.xctest")
+        
+        server.registerRoute(route) {
+            if let clientCode = XPCRequestContext.clientCode {
```
amomchilov, line 112: Use [`XCTUnwrap(_:_:file:line:)`](https://developer.apple.com/documentation/xctest/3380195-xctunwrap):

```swift
let clientCode = XCTUnwrap(XPCRequestContext.clientCode)
...
```

Tests/SecureXPCTests/Client & Server/XPCRequestContext Tests.swift
```diff
@@ -0,0 +1,131 @@
...
+                clientCodeNotNil.fulfill()
+                var staticCode: SecStaticCode?
+                SecCodeCopyStaticCode(clientCode, [], &staticCode)
+                var signingInfo: CFDictionary?
+                SecCodeCopySigningInformation(staticCode!, [], &signingInfo)
+                if let signingInfo = signingInfo as NSDictionary?,
+                   signingInfo[kSecCodeInfoIdentifier] as? String == "com.apple.xctest" {
+                    xctestIdentifier.fulfill()
```
amomchilov, line 120: You only need a single expectation, to ensure that this route handler was called at all.

Once that's assured, you can just use regular assertions here:

```swift
let signingInfo = XCTUnwrap(signingInfo = signingInfo as NSDictionary?)
XCTAssertEqual(signingInfo[kSecCodeInfoIdentifier] as? String, "com.apple.xctest")
````

(Same applies above)

amomchilov (approved):
Love it.

One note on terminology:

So far we called the incoming thing a "message", not a "request", is that right? So perhaps `XPCMessageContext` would make sense in that context (heh).

But actually, I think we should use `message` to refer exclusively to the payload attached to a request, and keep `XPCRequestContext`

jakaplan:
> So far we called the incoming thing a "message", not a "request", is that right? So perhaps `XPCMessageContext` would make sense in that context (heh).

Terminology is hard! My intention, which I may not have consistently succeeded at, is that "message" refers to the object being sent via the client to the server. This means there are two route types which have no message, but they still result in a server handler being called. Hence I chose "request" instead of "message", but it's true this introduces yet another noun. I also considered `XPCClientContext` as it's describing the client, but I thought that was somewhat confusing and could be confused as something that is accessible on/to the client.

Sources/SecureXPC/XPCRequestContext.swift
jakaplan: Oh, good to know! Will change.

Tests/SecureXPCTests/Client & Server/XPCRequestContext Tests.swift
jakaplan: Ah cool, didn't know that existed

jakaplan: Oh but it `throws`, and I need to use it inside of the handler's closure so it'd just get swallowed :(

Sources/SecureXPC/XPCRequestContext.swift
jakaplan: How would I be able to automatically determine which type of storage should be used?

jakaplan: I'm going to merge this, but happy to revisit this once I understand what you're suggesting.

amomchilov: Using the same `@available(macOS 10.15.0, *)`  (or but in the `if #available ...` runtime variant) checks that you already use here.

amomchilov:
>  is that "message" refers to the object being sent via the client to the server. This means there are two route types which have no message, but they still result in a server handler being called.

I think it would make more sense if the terminology such that all 4 route types send a *request*, and some of them have a *message* attached.

Then `XPCRequestContext` would make perfect sense as a name for this

Sources/SecureXPC/XPCRequestContext.swift
jakaplan: But setting a synchronous error handler is still possible (and perfectly reasonable) to use on macOS 10.15 or later.

Right now the caller explicitly calls either `setForTask(...)` or `setForCurrentThread(...)` which lets me know which type of storage needs to be used. I've been interpreting your comment as being about automatically determining whether the storage needs to be per task or per thread, but without an explicit decision I don't know how to determine that.

PR #74, opened 2022-02-23
Adds ability to use an async error handler with `XPCServer`

jakaplan:
Also added tests to validate that the server's error handler actually gets called.


jakaplan:
@amomchilov For your review. Unless I overlooked something (totally possible), this should complete adding `async` support within SecureXPC.

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -350,11 +350,52 @@ public class XPCServer {
...
+        
+        func handle(_ error: XPCError) {
+            switch self {
+                case .none:
+                    ()
+                case .sync(let handler):
+                    handler(error)
+                case .async(let handler):
+                    if #available(macOS 10.15, *) {
+                        Task {
```
amomchilov, line 369: Is spawning a new task the correct semantics here?

Sources/SecureXPC/Server/XPCServer.swift
```diff
@@ -350,11 +350,52 @@ public class XPCServer {
...
+    /// Wrapper around an error handling closure to ensure there's only ever one error handler regardless of whether it's synchronous or asynchronous.
+    fileprivate enum ErrorHandler {
+        case none
+        case sync((XPCError) -> Void)
+        case async((XPCError) async -> Void)
+        
+        func handle(_ error: XPCError) {
+            switch self {
+                case .none:
+                    ()
```
amomchilov, line 364: Doesn't this trigger an unused-value warning? This Void value isn't returned (implicit `return` only works for single-line expressions). Control flow falls out of here and falls to the end of the function normally.

Typically you would just use `break`, or in this case, just `return`.

Tests/SecureXPCTests/Client & Server/Server Error Handler Test.swift
```diff
@@ -0,0 +1,80 @@
...
+// Note: it's intentional this is *not* `Codable` to validate that server side all errors are sent to the error handler
+private enum ExampleError: Error, Equatable {
+    case completeAndUtterFailure
+}
+
+class ServerErrorHandlerTest: XCTestCase {
+    
+    func testErrorHandler_Sync()  throws {
+        let errorToThrow = ExampleError.completeAndUtterFailure
+        let errorExpectation = self.expectation(description: "\(errorToThrow) should be provided to error handler")
```
amomchilov, line 21: Even if you're not using async code in the SUT, you can use it in your test here to remove the need for expectations.

https://www.swiftbysundell.com/articles/unit-testing-code-that-uses-async-await/#an-alternative-to-expectations

amomchilov: NVM, I see why it's not so straight forward in this case. See my comments below.

Tests/SecureXPCTests/Client & Server/Server Error Handler Test.swift
```diff
@@ -0,0 +1,80 @@
...
+        
+        let failureRoute = XPCRoute.named("always", "throws")
+        let server = XPCServer.makeAnonymous()
+        let client = XPCClient.forEndpoint(server.endpoint)
+        server.registerRoute(failureRoute) {
+            throw ExampleError.completeAndUtterFailure
+        }
+        server.setErrorHandler { (error: XPCError) async -> Void in
```
amomchilov, line 60: Hmmmm I'm wondering if this closure-based error handling API makes sense in an in the new concurrency world.

I see the benefit they have w.r.t. centralizing error handling for common failures, but they can be a bit cumbersome to unwrap. I can't think of anything better, though.

amomchilov (approved):

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: I believe so. My understanding is if the current context is async then it'll attach to the TaskGroup under execution. Otherwise it'll cause a new one to be created. That behavior seems desirable.

Sources/SecureXPC/Server/XPCServer.swift
jakaplan: It does not trigger any warning. But `break` does indeed make more sense.

Tests/SecureXPCTests/Client & Server/Server Error Handler Test.swift
jakaplan: On the server side, I don't think the introduction of `async` changes much. Concurrent execution of server handlers was always easily achievable, it's just now possible for them to yield partway through their execution.

However, you bringing this up makes me wonder about few different alternatives: 
1. Each route can optionally register an error handler in addition to the current "operation" handler, which receives `XPCError`s just like the global handler does at present. There's no global handler provided by the framework, but API users can always register the same error handler for all routes.
2. The same as (1), but a global handler can set and receives errors for routes that don't have an error handler set.
3. Each route can optionally register an error handler in addition to the current "operation" handler, which receives `Error`s that it threw during execution. (Potentially also decoding errors for just the payload.) All other errors are sent to a global handler provided by the framework.

I quite like the idea of (3). Thoughts? We can also move this to a discussion.

amomchilov: Yeah, worth opening an issue/discussion. I don't currently have any particular thoughts though, because all my server routes do `try!` with a `// TODO: handle errors` :')

PR #75, opened 2022-02-24
Introduces optional error handlers for each route

jakaplan:
A global error handler remains, but now exists for global errors like a request for route that's not registered or insecure connection.

Decoding of a message type and encoding of a reply type and/or error type are still sent to the global error handler. Ideally those would also be sent to the route specific one. If we like this approach, can take a look at doing that next.

### DIFF
--- a/Sources/SecureXPC/Server/XPCServer.swift
+++ b/Sources/SecureXPC/Server/XPCServer.swift
@@ -80,15 +80,15 @@ import Foundation
 /// - ``makeAnonymous()``
 /// - ``makeAnonymous(clientRequirements:)``
 /// ### Registering Routes
-/// - ``registerRoute(_:handler:)-4ttqe``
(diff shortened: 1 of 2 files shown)


jakaplan:
@amomchilov I'd appreciate getting your thoughts on this approach. If you think it's a good idea, I'd also welcome a code review although this approach is currently incomplete - it should also have applicable decoding/encoding errors go to the route's error handler and not the global one.

jakaplan:
@amomchilov Thought on this approach? Genuinely unsure if this is actually helpful or not.

jakaplan:
Not convinced this is the correct approach, so closing for the time being. Can always be reconsidered in the future.

amomchilov:
Sorry, this slipped through the cracks, I never even took a look at it :o

PR #77, opened 2022-02-26
Renames `route` parameter to `toRoute` for `send` functions

jakaplan:
This is done to align with Swift's API Guidelines that state, "Prefer method and function names that make use sites form grammatical English phrases."

The `toRoute` parameter is always the first or second parameter, which makes it quite important. As the guidelines further state, "It is acceptable for fluency to degrade after the first argument or two when those arguments are not central to the call’s meaning."

There are no functional changes to this PR, just a naming change. Of course this is a breaking change.

### DIFF
--- a/Sources/SecureXPC/Client/XPCClient.swift
+++ b/Sources/SecureXPC/Client/XPCClient.swift
@@ -50,14 +50,14 @@ import Foundation
 /// Once a client has been retrieved, sending a request is as simple as invoking `send` with a route:
 /// ```swift
 /// let resetRoute = XPCRoute.named("reset")
-/// await client.send(route: resetRoute)
+/// try await client.send(toRoute: resetRoute)
 /// ```
 ///
 /// If the client needs to receive information back from the server, a route with a reply type must be used:
 /// ```swift
 /// let currentRoute = XPCRoute.named("config", "current")
 ///                            .withReplyType(Config.self)
-/// let config = try await client.send(route: currentRoute)
+/// let config = try await client.send(toRoute: currentRoute)
 /// ```
 ///
 /// Routes can also require a message be included in the request:
@@ -66,7 +66,7 @@ import Foundation
 ///                           .withMessageType(Config.self)
 ///                           .withReplyType(Config.self)
 /// let config = <# create Config instance #>
-/// let updatedConfig = try await client.sendMessage(config, route: updateRoute)
+/// let updatedConfig = try await client.sendMessage(config, toRoute: updateRoute)
 /// ```
 ///
 /// ### Sending Requests with Closures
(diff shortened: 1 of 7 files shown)


jakaplan:
@amomchilov Just wanted to flag this PR to you since it's a breaking change. Otherwise this PR isn't interesting and so I'm going to merge it without review.

Sources/SecureXPC/Client/XPCClient.swift
```diff
@@ -210,11 +212,13 @@ public class XPCClient {
     /// Sends a request with no message that does not receive a reply.
     ///
     /// - Parameters:
-    ///   - route: The server route which will handle this request.
+    ///   - toRoute: The server route which will handle this request.
     @available(macOS 10.15.0, *)
-    public func send(route: XPCRouteWithoutMessageWithoutReply) async throws {
+    public func send(
+        toRoute route: XPCRouteWithoutMessageWithoutReply
```
amomchilov, line 218: I would argue that this should be:

```suggestion
        to destination route: XPCRouteWithoutMessageWithoutReply
```

* *Omit needless words.* Every word in a name should convey salient information at the use site.

    More words may be needed to clarify intent or disambiguate meaning, but those that are redundant with information the reader already possesses should be omitted. In particular, omit words that merely repeat type information.

    ```diff
    - public mutating func removeElement(_ member: Element) -> Element?
    
    - allViews.removeElement(cancelButton)
    ```
    In this case, the word Element adds nothing salient at the call site. This API would be better:

    ```diff
    + public mutating func remove(_ member: Element) -> Element?
    
    + allViews.remove(cancelButton) // clearer
    ```
    
* *Name variables, parameters, and associated types according to their roles*, rather than their type constraints.

    ```diff
    - var string = "Hello"
    - protocol ViewController {
    -  associatedtype ViewType : View
    - }
    - class ProductionLine {
    -   func restock(from widgetFactory: WidgetFactory)
    - }
    ```

    Repurposing a type name in this way fails to optimize clarity and expressivity. Instead, strive to choose a name that expresses the entity’s role.
    
    ```diff
    + var greeting = "Hello"
    + protocol ViewController {
    +   associatedtype ContentView : View
    + }
    + class ProductionLine {
    +  func restock(from supplier: WidgetFactory)
    + }
    ```

jakaplan: Thanks for the additional context. That makes sense. However, I'm confused by your suggestion:
```swift
to destination route: XPCRouteWithoutMessageWithoutReply
```

As far I'm aware (and I double checked in Xcode) this doesn't appear to be valid? Specifically the middle `destination` token isn't valid syntax. Or am I missing something?

amomchilov: oops, I meant:

```suggestion
        to destination: XPCRouteWithoutMessageWithoutReply
```

PR #97, opened 2022-06-27
Partial support for macOS 13's new `SMAppService`

jakaplan:
- Adds support for daemons and this has been tested
 - Updates documentation for login items, which _may_ just work, but it's actually unclear because Apple's documentation right now is obviously incorrect. See https://developer.apple.com/forums/thread/709171#709171021
 - No support yet for launch agents

However, this is making the API surface increasing unwieldy so I have thoughts on how to refactor this to in most cases auto-detect which type of server  (and for that matter client as well) and let users explicitly specify a "type" if so desired. But first, wanted to land this change in the current style (although likely not make a release containing it).

### DIFF
--- a/Sources/SecureXPC/Server/XPCMachServer.swift
+++ b/Sources/SecureXPC/Server/XPCMachServer.swift
@@ -128,43 +128,32 @@ extension XPCMachServer {
         // Determine mach service name using the launchd property list's MachServices entry
         let machServiceName: String
         let launchdData = try readEmbeddedPropertyList(sectionName: "__launchd_plist")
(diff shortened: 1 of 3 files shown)


jakaplan:
FYI @amomchilov going to work on a number of changes in this area to both support macOS 13's new `SMAppService` and simplifying SecureXPC's API surface for creating servers and clients. This doesn't do the latter, I'm planning to put up another PR soonish (haven't started it yet) to do that.

jakaplan:
Abandoning this in favor of a new pull request

amomchilov:
For future readers, this was implemented in https://github.com/trilemma-dev/SecureXPC/pull/99

PR #99, opened 2022-07-01
Adds support for `SMAppService` agents & daemons

jakaplan:
Implemented and tested on macOS 13.

### DIFF
--- a/Sources/SecureXPC/SecureXPC.docc/SecureXPC.md
+++ b/Sources/SecureXPC/SecureXPC.docc/SecureXPC.md
@@ -8,11 +8,16 @@ compatability.
 SecureXPC provides an easy way to perform secure XPC communication with pure Swift. `Codable` conforming types are used
 to make requests and receive responses.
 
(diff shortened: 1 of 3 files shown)


jakaplan:
@amomchilov available for you to review if you'd like. This is to add support for the new agents & daemons functionality as part of [`SMAppService`](https://developer.apple.com/documentation/servicemanagement/smappservice). The daemon functionality is conceptually (but not in implementation) a replacement for `SMJobBless`.

amomchilov:
Interesting, I'm learning about SMAppService from you for the first time. Was there a WWDC video on it? I can't find anything.

jakaplan:
> Was there a WWDC video on it? I can't find anything.

Nope, nor a sample. The documentation is present, but not great. I was able to use it enough to install a daemon and an agent, but I have a feeling I'm not using it quite properly. The login item documentation for it is obviously wrong and nonsensical; I asked a question on the Apple Developer Forums a few weeks ago and have not gotten a response which is unlike Eskimo.

PR #101, opened 2022-07-05
`XPCClientRequirement` now used internally

jakaplan:
`MessageAcceptor` still exists, but is purely an internal implementation details of `XPCClientRequirement` that nowhere else in the codebase knows about (nor has access to).

This means that API users can create requirements just like SecureXPC does, with two intentional carve outs:
 - `alwaysAccepting` which is internally only ever used as the default for an `XPCServiceServer`, using it elsewhere defeats the security of this package (and if such functionality is truly desired, it can be accomplished by using a highly permissive `SecRequirement` - but the fact doing so is cumbersome is a feature, not a bug)
- `sameProcess` which is internally only ever used as the default for an `XPCAnonymousServer` which is safe and highly convenient in that exact use case, but is absolutely insecure for a Mach service as its prone to PID races

`XPCClientRequirement` has been changed from an enum to a struct with identical names which means it can be used the same at the call site except that now it throws at access time which I think is a much clearer way of representing failure.


jakaplan:
@amomchilov available for your review if you'd like to (and post-merge reviews are also always most welcome)

Sources/SecureXPC/Server/XPCClientRequirement.swift
```diff
@@ -8,88 +8,364 @@
...
+        // In the case of an XPCAnonymousServer, all of the connections must be created after the server itself was
+        // created. As such, the process containing the server must always exist first and so no other process can
+        // have the same PID while that process is still running. While it's possible the process now corresponding to
+        // the PID returned by xpc_connection_get_pid(...) is not the process that created the connection, there's no
+        // way for it fake being this process. Therefore for anonymous connections it's safe to directly compare PIDs.
+        getpid() == xpc_connection_get_pid(connection)
+    }
+    
+    func isEqual(to acceptor: MessageAcceptor) -> Bool {
+        type(of: acceptor) == SameProcessMessageAcceptor.self
```
amomchilov, line 193:
```suggestion
    acceptor is SameProcessMessageAcceptor
```

Sources/SecureXPC/Server/XPCClientRequirement.swift
```diff
@@ -8,88 +8,364 @@
...
+        }
+        
+        // Validate the each component in the client path is present in the parent bundle path in the same order
+        for i in 0..<parentBundleURL.pathComponents.count {
+            if parentBundleURL.pathComponents[i] != clientPath.pathComponents[i] {
+                return false
+            }
+        }
+        
+        return true
```
amomchilov, line 282: You can replace this manual indexing with `zip`, and this condition checking with `allSatisfy(_:)` (in other cases where you conditionally `return true` else `return false`, you can use `contains(where:)`)

```suggestion
        return zip(parentBundleURL.pathComponents, clientPath.pathComponents).allSatisfy(==)
```

Sources/SecureXPC/Server/XPCClientRequirement.swift
```diff
@@ -8,88 +8,364 @@
...
+            return nil
+        }
+        
+        var path: CFURL?
+        guard Security.SecCodeCopyPath(staticCode, SecCSFlags(), &path) == errSecSuccess,
+              let path = (path as URL?)?.standardized else {
+            return nil
+        }
+        
+        return path
```
amomchilov, line 298: This `path` shadowing is tricky, but not actually necessary here

```suggestion
        guard Security.SecCodeCopyPath(staticCode, SecCSFlags(), &path) == errSecSuccess else {
            return nil
        }
        
        return (path as URL?)?.standardized
```

Sources/SecureXPC/Server/XPCClientRequirement.swift
```diff
@@ -8,88 +8,364 @@
...
+        guard let acceptor = acceptor as? OrMessageAcceptor else {
+            return false
+        }
+        
+        return self.lhs.isEqual(to: acceptor.lhs) && self.rhs.isEqual(to: acceptor.rhs)
+    }
+}
+
+/// If sandboxed, expands the sandbox by using a URL contained within the client's request.
+private func expandSandboxIfNecessary(message: xpc_object_t) throws {
```
amomchilov, line 359: I'm so happy that you know all of this niche information, because I sure don't~ (until now!)

Sources/SecureXPC/Server/XPCClientRequirement.swift
```diff
@@ -8,88 +8,364 @@
...
-    case and(XPCClientRequirement, XPCClientRequirement)
+    public static func and(_ lhs: XPCClientRequirement, _ rhs: XPCClientRequirement) -> XPCClientRequirement {
+        XPCClientRequirement(messageAcceptor: AndMessageAcceptor(lhs: lhs.messageAcceptor, rhs: rhs.messageAcceptor))
+    }
     
     /// The requesting client must satisfy at least one of the requirements.
-    case or(XPCClientRequirement, XPCClientRequirement)
+    public static func or(_ lhs: XPCClientRequirement, _ rhs: XPCClientRequirement) -> XPCClientRequirement {
+        XPCClientRequirement(messageAcceptor: OrMessageAcceptor(lhs: lhs.messageAcceptor, rhs: rhs.messageAcceptor))
+    }
```
amomchilov, line 113: I think it's be pretty nice/natural/expressive to use `&&` and `||` operators for this. It would let people have bracket-less, infix notatio

Like: `a && b && c && d`
Instead of: `and(a, and(b, and(c, d)`


Alternatively, `and` `or` could be replaced with `all()` and `any`. This might have some performance benefit w.r.t. allocating fewer existential containers as compared to a tree of `AndMessageAcceptor`/`OrMessageAcceptor` instances.

Sources/SecureXPC/Server/XPCMachServer.swift
```diff
@@ -368,9 +355,9 @@ extension XPCMachServer {
             return false
         }
         
-        // Login item must be within the main app bundle's Contents/Library/LoginItems directory
+        // Login item must be within the main bundle's Contents/Library/LoginItems directory
         let pathComponents = loginItem.deletingLastPathComponent().pathComponents
-        guard pathComponents.count >= 3, pathComponents.suffix(3) == ["LoginItems", "Library", "Contents"] else {
+        guard pathComponents.count >= 3, pathComponents.suffix(3) == ["Contents", "Library", "LoginItems"] else {
```
amomchilov, line 360: heh, nice catch.

Sources/SecureXPC/Server/XPCMachServer.swift
```diff
@@ -181,32 +181,31 @@ extension XPCMachServer {
...
+            guard SecRequirementCreateWithString(client as CFString, SecCSFlags(), &requirement) == errSecSuccess,
+                  let requirement = requirement else {
+                throw XPCError.misconfiguredServer(description: "Invalid SMAuthorizedClients requirement: \(client)")
+            }
+            
+            if let currentRequirement = clientRequirement {
+                clientRequirement = .or(currentRequirement, .secRequirement(requirement))
+            } else {
+                clientRequirement = .secRequirement(requirement)
+            }
```
amomchilov, line 202: This is essentially building up a linked list. Would an array be preferable? It'd be simpler to construct (just a simple `compactMap`), simpler to check `contains(where: $0.acceptsMessage(...))`, and might be nicer on allocations/memory.

The math nerd in me is compelled to mention that this is called "existential quantification" ("any"), and the opposite is "universal quantification" ("all")

Sources/SecureXPC/Server/XPCServiceServer.swift
```diff
@@ -118,20 +118,15 @@ internal class XPCServiceServer: XPCServer {
...
-            if let teamID = try? teamIdentifierForThisProcess(),
-               let teamIDAcceptor = try? SecRequirementsMessageAcceptor(forTeamIdentifier: teamID) {
-                self.messageAcceptor = AndMessageAcceptor(lhs: teamIDAcceptor, rhs: parentBundleAcceptor)
+            // Now that any arbitrary process could create a connection to this server, the client requirement needs to
+            // be more restrictive. We'll allow any connection from a process belongs to the same parent bundle and if
+            // there's a valid team ID present we'll additionally enforce it's of the same team ID. Creating the parent
+            // bundle requirement should always succeed because as part of creating an XPCServiceServer a check is
+            // performance that this process is located within a Contents/XPCServices directory.
+            if let teamIDRequirement = try? XPCClientRequirement.sameTeamIdentifier {
+                self.clientRequirement = .and(try! .sameParentBundle, teamIDRequirement)
```
amomchilov, line 127: Nice!

Sources/SecureXPC/Server/XPCClientRequirement.swift
```diff
@@ -8,88 +8,364 @@
...
+}
+
+/// This should only be used by XPC services which are by default application-scoped, so it's acceptable to assume they're inheritently safe.
+fileprivate struct AlwaysAcceptingMessageAcceptor: MessageAcceptor {
+    func acceptMessage(connection: xpc_connection_t, message: xpc_object_t) -> Bool {
+        true
+    }
+    
+    func isEqual(to acceptor: MessageAcceptor) -> Bool {
+        type(of: acceptor) == AlwaysAcceptingMessageAcceptor.self
```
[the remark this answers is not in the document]
jakaplan: Didn't even know about the `is` keyword!

Sources/SecureXPC/Server/XPCClientRequirement.swift
```diff
@@ -8,88 +8,364 @@
...
+    }
+}
+
+// MARK: Internal implementation
+
+// A `MessageAccceptor` is essentially the internal implementation for an `XPCClientRequirement`
+
+fileprivate protocol MessageAcceptor {
+    /// Determines whether an incoming message should be accepted.
+    func acceptMessage(connection: xpc_connection_t, message: xpc_object_t) -> Bool
```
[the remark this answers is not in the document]
jakaplan: That seems nice, but in the context of Swift I feel like would be confusing with optionals. I've on rare occasion confused myself when reading ternary operators.

Sources/SecureXPC/Server/XPCClientRequirement.swift
```diff
@@ -8,88 +8,364 @@
...
+
+/// This is intended for use by `XPCAnonymousServer`.
+fileprivate struct SameProcessMessageAcceptor: MessageAcceptor {
+    /// Accepts a message only if it is coming from this process.
+    func acceptMessage(connection: xpc_connection_t, message: xpc_object_t) -> Bool {
+        // In the case of an XPCAnonymousServer, all of the connections must be created after the server itself was
+        // created. As such, the process containing the server must always exist first and so no other process can
+        // have the same PID while that process is still running. While it's possible the process now corresponding to
+        // the PID returned by xpc_connection_get_pid(...) is not the process that created the connection, there's no
+        // way for it fake being this process. Therefore for anonymous connections it's safe to directly compare PIDs.
```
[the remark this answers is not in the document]
jakaplan: If I instead wrote:
> Therefore for connections received by an XPCAnonymousServer it's safe to directly compare PIDs.

Would you find the above an accurate description? My intention in writing this wasn't actually about anonymous connections, it was about how `XPCAnonymousServer` behaves in regards to receiving connections.

Sources/SecureXPC/Server/XPCClientRequirement.swift
jakaplan: Interesting, I didn't realize `zip` could handle inputs of two different lengths and that it would handle the shorter one. That's quite convenient, although I do think in some ways it makes the code a little harder to reason about due to need to remember that aspect of it. Nonetheless, will switch over to using it.

Sources/SecureXPC/Server/XPCMachServer.swift
jakaplan: Yup 😬

Wish there was a good way to test these in automated way, but short of a sophisticated integration test I don't think it's going to happen. And setting up and maintaining that seems prohibitive.

Sources/SecureXPC/Server/XPCClientRequirement.swift
jakaplan: Ah yeah that could be nice, hadn't thought of that. Let me try it out.

jakaplan: Yeah the custom operator implementations are nice, will change to using them.

Sources/SecureXPC/Server/XPCClientRequirement.swift
jakaplan: Well I only known it now through reading lots of stuff, asking some questions, and some trial and error. XPC is rather arcane.

Sources/SecureXPC/Server/XPCClientRequirement.swift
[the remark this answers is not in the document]
amomchilov: It's marginally different semantics, in that is considers subclass relationships as well. `type(_:)`+`==` approach is still useful to know, it's how you can do exact type checks.

Check out this answer of mine for some more details: https://stackoverflow.com/a/40388434/3141234

Sources/SecureXPC/Server/XPCClientRequirement.swift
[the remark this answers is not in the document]
amomchilov: And the `?()` syntax is already "taken" for calling closures of optional type or optional methods on Objective C protocols.

Sources/SecureXPC/Server/XPCClientRequirement.swift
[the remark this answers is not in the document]
amomchilov: Ah, yes, I think that makes sense.

Sources/SecureXPC/Server/XPCClientRequirement.swift
amomchilov: Indeed. I think that's relative well known, so it's safe to rely on.

The less obvious part is what to do if you want to have it pair up to the length of the longer sequence (e.g. pairing the elements of the longer sequence with `nil`). There's no built in solution for that.

PR #106, opened 2022-07-10
Property wrappers for `IOSurface` and file handle/descriptor

jakaplan:
Adds the ability to send `IOSurface` instance across an XPC connection. Repurposes the "manual" wrapper previously implemented for file handle/descriptor such that it now is three different property wrappers that all use the same underlying implementation.


jakaplan:
@amomchilov If you're up for reviewing, I'd be much appreciative. I don't have much experience with creating property wrappers and so would benefit from your expertise in this area.

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
```diff
@@ -0,0 +1,184 @@
...
+                                                debugDescription: "Encoding failed for \(self.descriptor)",
+                                                underlyingError: nil)
+            throw EncodingError.invalidValue(self.descriptor, context)
+        }
+        
+        if self.closeOnEncode {
+            try self.close()
+        }
+        container.encode(encodedDescriptor, forKey: CodingKeys.descriptor)
+        try container.encode(self.closeOnEncode, forKey: CodingKeys.closeOnEncode)
```
amomchilov, line 39: This is a weird situation. We failed to close it, but we're going to try to encode it anyway. Not sure what the "safe" semantics should be here.

Tests/SecureXPCTests/Encoder & Decoder/Round Trip/File Descriptor XPC Wrapper Tests.swift
```diff
@@ -0,0 +1,194 @@
...
+final class FileDescriptorXPCContainerTests: XCTestCase {
+    // MARK: helper functions
+    
+    private func currentPath(filePath: String = #filePath) -> String { filePath }
+    
+    private func pathForFileDescriptor(fileDescriptor: Int32) -> String {
+        var buffer = [CChar](repeating: 0, count: Int(PATH_MAX))
+        _ = fcntl(fileDescriptor, F_GETPATH, &buffer)
+        
+        return String(cString: buffer)
```
amomchilov, line 22: Wow I'm baffled to learn that [`Foundation.FileHandle`](https://developer.apple.com/documentation/foundation/filehandle) doesn't have an API for this.

I was hoping to avoid dealing with overly terse C gobbledygook, but looks like that's the only option here!

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
```diff
@@ -0,0 +1,184 @@
...
+///
+/// By default the provided file descriptor will be closed once it has been encoded.
+///
+/// When creating an ``XPCRoute`` that directly transfers this type as either the message or reply type, `DarwinFileDescriptorForXPC` must be the
+/// specified type, not `CInt`. This is not applicable when transferring a type which _contains_ a wrapped file descriptor as one of its properties.
+///
+/// This property wrapper shares an underlying representation with ``FileHandleForXPC`` and ``FileDescriptorForXPC`` and therefore may be used
+/// interchangeably between the server and client. However, to make use of such functionality requires routes with identical names and differing message and/or
+/// reply types.
+@propertyWrapper public struct DarwinFileDescriptorForXPC {
```
amomchilov, line 73: What's a "Darwin file descriptor", as compared to a "normal" file descriptor of unix fame?

amomchilov: NVM the darwin one *is* the unix one, whereas FileDescriptor is from the system framework. https://developer.apple.com/documentation/system/filedescriptor

Got it.

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
```diff
@@ -0,0 +1,184 @@
...
+/// By default the provided file descriptor will be closed once it has been encoded.
+///
+/// When creating an ``XPCRoute`` that directly transfers this type as either the message or reply type, `FileDescriptorForXPC` must be the specified type,
+/// not `FileDescriptor`. This is not applicable when transferring a type which _contains_ a wrapped file descriptor as one of its properties.
+///
+/// This property wrapper shares an underlying representation with ``DarwinFileDescriptorForXPC`` and ``FileHandleForXPC`` and therefore may be
+/// used interchangeably between the server and client. However, to make use of such functionality requires routes with identical names and differing message and/or
+/// reply types.
+@available(macOS 11.0, *)
+@propertyWrapper public struct FileDescriptorForXPC {
```
amomchilov, line 118: Would it be more explicit to wrap this with a `#if canImport(System)`? Not sure the pros/cons.

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
```diff
@@ -0,0 +1,184 @@
...
+///
+/// By default the provided file handle will be closed once it has been encoded.
+///
+/// When creating an ``XPCRoute`` that directly transfers this type as either the message or reply type, `FileHandleForXPC` must be the specified type, not
+/// `FileHandle`. This is not applicable when transferring a type which _contains_ a wrapped file handle as one of its properties.
+///
+/// This property wrapper shares an underlying representation with ``DarwinFileDescriptorForXPC`` and ``FileDescriptorForXPC`` and therefore
+/// may be used interchangeably between the server and client. However, to make use of such functionality requires routes with identical names and differing
+/// message and/or reply types.
+@propertyWrapper public struct FileHandleForXPC {
```
amomchilov, line 157: How do you feel about using `Foundation.FileHandle` as the fundamental file-descriptor abstraction, instead of a property-wrapped plain `Int32`/`CInt`?

I think we should be opinionated and only support one of the two, and of the two, I think `FileHandle` is better.

I think this has two upsides and one downside:

1. Pro: the wrapped type is more specific. It's a specific struct, rather than a general integer.
2. Pro: the type already exists, and has `closeOnDealloc` just like `cloneOnEncode`. We don't need to keep our own bool or do our own close logic.
3. Con: It makes `Foundation` a required dependency, but I don't think that matters given that we already require XPC and are necessarily on Apple platforms that have Foundation.

Furthermore: I would just suggest simply doing `extension FileHandle: FileDescriptorCodable`, and removing the complexity of a property wrapper from this API design. It's perfectly safe to do this kind of retroactive modelling, as long as you're the authority over either the conforming type (We're not, Apple own `Foundation.FileHandle`) or the protocol (we are)

Tests/SecureXPCTests/Encoder & Decoder/Round Trip/File Descriptor XPC Wrapper Tests.swift
```diff
@@ -0,0 +1,194 @@
...
+        let document = try await client.send(to: route).document
+        defer { close(document) }
+        
+        XCTAssertEqual(pathForFileDescriptor(fileDescriptor: document), currentPath())
+    }
+    
+    // MARK: Automatic bridging
+    
+    func testAutomaticBridging_DirectInit() async throws {
+        let server = XPCServer.makeAnonymous()
```
amomchilov, line 148: These anonymous services are coming in really clutch for testing! I'm so glad I ran into that one random forum post that suggested it lol

Tests/SecureXPCTests/Encoder & Decoder/Round Trip/File Descriptor XPC Wrapper Tests.swift
```diff
@@ -0,0 +1,194 @@
...
+
+import System
+import Foundation
+import XCTest
+import SecureXPC
+
+final class FileDescriptorXPCContainerTests: XCTestCase {
+    // MARK: helper functions
+    
+    private func currentPath(filePath: String = #filePath) -> String { filePath }
```
amomchilov, line 16: Clever!

amomchilov (commented):
I dig it!

Left a few comments about slimming down the FD related APIs, but I think the overall feature is well worthwhile and well implemented!

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
jakaplan: If it fails to close, shouldn't that cause an error to be thrown and therefore the encoding won't happen?

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
jakaplan: Yeah that was the best name I could think of. If you have alternative recommendations I'd be happy to hear them. The other one I though of was "native" but preferred "Darwin" to that.

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
jakaplan: It was introduced in macOS 11 as far as I can tell from the documentation, is this not the normal way to conditionally handle this on macOS? I've never done `canImport` for a package/framework that ships with an Apple OS (just optional 3rd party dependencies).

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
jakaplan: Let me think about it and will reply with more thoughts later.

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
amomchilov: Oh right, true.

What about the result of `xpc_fd_create`? The docs say it acts like `dup()` was called on the parameter. If an error is thrown here, should we catch it, close the `encodedDescriptor`, then rethrow the error?

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
amomchilov: > is this not the normal way to conditionally handle this on macOS?

I'm not sure!

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
jakaplan: My interpretation is it's _like_ `dup()`, but not actually the same in regard to the resulting `xpc_object_t` (or something opaquely referenced by it) needs to be explicitly closed. I say that because it appears to be impossible to close it. Once it's been encoded as an `xpc_object_t` the only way to get back a file descriptor is to call `xpc_fd_dup()` which itself says:

> Returns a file descriptor that is equivalent to the one boxed by the file file descriptor object.

and also says:

> Multiple invocations of xpc_fd_dup() will not return the same file descriptor number, but they will return descriptors that are equivalent, as though they had been created by dup(2).

My reading of it is an attempt to get a file descriptor using `xpc_fd_dup()` in order to close it would be pointless as it's `dup()`ing it each time one is requested - it's not a "true" unboxing of the original value.

Tests/SecureXPCTests/Encoder & Decoder/Round Trip/File Descriptor XPC Wrapper Tests.swift
jakaplan: Yeah, I was expecting `System.FileDescriptor` to have an API for this since it seems like it's meant to be a nice object oriented wrapper over C functionality.

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
jakaplan: POSIX sounds good!

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
jakaplan: The one I have there was autosuggested by Xcode so I'll stick with it.

Sources/SecureXPC/Codable Property Wrappers/File Descriptor XPC Wrappers.swift
jakaplan: > How do you feel about using Foundation.FileHandle as the fundamental file-descriptor abstraction, instead of a property-wrapped plain Int32/CInt?

Do you mean internally or that there be no publicly exposed API for using POSIX file descriptors?

> Pro: the type already exists, and has `closeOnDealloc` just like `cloneOnEncode`. We don't need to keep our own bool or do our own close logic.

---

I'd like to support `System.FileDescriptor` as well as it doesn't appear to have a `closeOnDealloc` equivalent as far as I can tell. And it goes as far as to say:
> You are responsible for managing the lifetime and validity of FileDescriptor values, in the same way as you manage a raw C file handle.

---

> I would just suggest simply doing `extension FileHandle: FileDescriptorCodable`, and removing the complexity of a property wrapper from this API design. 

How would one do that? I tried this a while back directly with `Decodable` (which ought to be an equivalent situation), but it appeared not to be possible because it has a required initializer which can't be added via an extension:

> 'required' initializer must be declared directly in class 'FileHandle' (not in an extension)

amomchilov: > Do you mean internally or that there be no publicly exposed API for using POSIX file descriptors?

no publicly exposed API for using POSIX file descriptors.

> I'd like to support System.FileDescriptor as well as it doesn't appear to have a closeOnDealloc equivalent as far as I can tell.

True, so we'll need one type for `Foundation.FileHandle` and one for `System.FileDescriptor`

> but it appeared not to be possible because it has a required initializer which can't be added via an extension

Ah... that might be a show stopper for that idea.

PR #110, opened 2022-07-13
Addresses misc feedback

jakaplan:
Addresses feedback for several recent PRs, not including the property wrapper one. There's also a change in here to make `XPCServer` be consistent with `XPCClient` in making it clear *what* type of requirement is needed (now that are two different kinds).


jakaplan:
@amomchilov For your reviewing pleasure.

FYI I'm about to take a long weekend to go skiing, TBD whether I'm going to take my laptop with me - so might be non-responsive until next week. At which point I'll go reengage with property wrapper PR.

amomchilov (approved):

amomchilov:
@jakaplan  Enjoy your trip, and thanks for patching all this feedback in!

PR #113, opened 2022-07-18
More efficient representations of `Data` and some array types

### LINKED ISSUE #7: Improve `Data` encoding efficiency by using its native XPC representation
> (That said, does it actually end up as a string?).

It's *much* worse than that. Storing it as a string (with say, Base64 for example), would have a 1.6x overhead (5 bits of payload per 8 bit character). The current encoding has an ***8x*** overhead. It encodes one `UInt8` of payload per `UInt64` of payload.

I think we should support the native `xpc_data` type

I think you missed a section of [one of my long comments](https://github.com/trilemma-dev/SecureXPC/pull/6#issuecomment-960865476). I quoted it below

> I think this is easy enough to do that it might be worth doing.
> 
> In particular, I've noticed that `Data` is ***really*** inefficient to encode. As it is today, it bloats sizes by 8x. It tries to encode itself as a `[UInt8]`, but each byte gets widened into an 8 byte `int64` in order to be passed to `xpc_int64_create`. Luckily, there is not per-number object allocation overhead there (that would have been brutal for my use-case haha), because the only possible int64 values here would be 0-255, all of which fit into the tagged pointer representation.
> 
> You can get a rough of idea of what that might look like, here: https://github.com/amomchilov/SecureXPC/compare/add-intial-tests-for-XPCEncoder...Alex/new-scalar-support . Of course, I only changed the encoder half, and the decoder would need to match.

_Originally posted by @amomchilov in https://github.com/trilemma-dev/SecureXPC/pull/6#discussion_r743985682_

jakaplan:
Resolves #7 by directly encoding/decoding `Data` to its XPC equivalent when it's the route reply or message type or enables an equally efficient encoding via an optional property wrapper.

The same is done for arrays of `Bool` and all standard numeric types.

### DIFF
--- a/Sources/SecureXPC/Codable Property Wrappers/ArrayOptimizedForXPC.swift
+++ b/Sources/SecureXPC/Codable Property Wrappers/ArrayOptimizedForXPC.swift
@@ -1,0 +1,136 @@
+//
+//  ArrayOptimizedForXPC.swift
+//  SecureXPC
+//
+//  Created by Josh Kaplan on 2022-07-19
+//
+
+import Foundation
+
+/// Wraps an array to optimize how it is sent over an XPC connection.
+///
+/// Arrays of the following type are supported by this property wrapper: `Bool`, `Double`, `Float`, `UInt`, `UInt8`, `UInt16`, `UInt32`, `UInt64`,
+/// `Int`, `Int8`, `Int16`, `Int32`, and `Int64`.
+///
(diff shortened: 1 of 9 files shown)


jakaplan:
@amomchilov for your review if you'd like. Due to how encoding/decoding works I'm pretty sure there's no way to just automatically make this work without API user intervention, so this is the best I could come up with. For the most basic case it will "just work" otherwise a property wrapper needs to be used which ought to be rather easy for the developer to do.

jakaplan:
I'm going to merge this in to continue work on some other changes, but always open to feedback at a later date

PR #116, opened 2022-07-21
Introduces shared memory capabilities

jakaplan:
The `SharedTrivial` property wrapper conforms to `Codable` but is not serializing its boxed value, it's actually mapping it into shared memory which can be accessed and modified across processes.

Note - I didn't write any tests for this because they unfortunately don't seem very useful as our tests all run in the same process. I could write some very basic ones, but they're not going to catch actual cross-process issues. I did test this a fair bit using an app with an XPC service.


jakaplan:
@amomchilov Would appreciate your eyes on this if you have time. Shared memory is not an area I'm an expert in, so definitely possible I got something wrong here.

Heads up my intention is to follow this up with a to-be-written PR for a collection like class (think a fixed capacity array of trivial types) and also potentially a fixed capacity `Data` like class.

jakaplan:
@amomchilov If you do end up seeing this PR, don't bother looking at it - just skip to https://github.com/trilemma-dev/SecureXPC/pull/119 instead.

PR #117, opened 2022-07-23
Dynamically creates a launch agent to facilitate testing

jakaplan:
A launch agent running as a completely separate process is now created, registered (load) with `launchd`, and can be communicated with. This allows for testing shared memory functionality in a separate process.

Cleanup logic also exists to unload the agent and delete all of the created files. However, the paths and service name are all randomly generated, so nothing bad ought to happen if cleanup fails (for example because a test fails).

Building SecureXPC from scratch takes a long time, so the code has been written in such a way that people using this framework can provide a path to the pre-built module - this will speed things up about 25x.

For reference, these were helpful resources in figuring out how to use `swiftc`:
- https://theswiftdev.com/building-static-and-dynamic-swift-libraries-using-the-swift-compiler/
- https://blog.spencerkohan.com/building-swift-without-a-build-system/

### DIFF
--- a/Tests/SecureXPCTests/Multiprocess Client & Server/LaunchAgent Tests.swift
+++ b/Tests/SecureXPCTests/Multiprocess Client & Server/LaunchAgent Tests.swift
@@ -1,0 +1,72 @@
+//
+//  LaunchAgentTests.swift
+//  SecureXPC
+//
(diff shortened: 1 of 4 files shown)


jakaplan:
@amomchilov For your future reference, if you want to write tests which access a launch agent that runs in a completely separate process that will now be possible.

PR #120, opened 2022-10-14
Added a new XPCServer.ClientRequirement for hardened runtime.

vincent-le-normand:
The goal of this PR is to validate that the client of a XPCServer has hardened runtime enabled.
This is in order to prevent some code injection security issues that existed prior to hardened runtime.

### DIFF
--- a/Sources/SecureXPC/Server/ClientRequirement.swift
+++ b/Sources/SecureXPC/Server/ClientRequirement.swift
@@ -60,6 +60,13 @@ extension XPCServer.ClientRequirement {
             return try teamIdentifier(teamID)
         }
     }
+
+	/// The requesting client must have hardened runtime enabled.
+	public static var hardenedRuntime: XPCServer.ClientRequirement {
+		get {
+			return XPCServer.ClientRequirement(messageAcceptor:HardenedMessageAcceptor())
+		}
+	}
     
     /// The requesting client must be within the same parent bundle as this server.
     ///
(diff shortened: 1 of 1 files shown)


Sources/SecureXPC/Server/ClientRequirement.swift
```diff
@@ -60,6 +60,13 @@ extension XPCServer.ClientRequirement {
             return try teamIdentifier(teamID)
         }
     }
+
+	/// The requesting client must have hardened runtime enabled.
```
jakaplan, line 64: From looking at [Apple's documentation](https://developer.apple.com/documentation/security/hardened_runtime) it appears they always refer to this as a proper noun capitalized as "Hardened Runtime".

Sources/SecureXPC/Server/ClientRequirement.swift
```diff
@@ -60,6 +60,13 @@ extension XPCServer.ClientRequirement {
...
+	public static var hardenedRuntime: XPCServer.ClientRequirement {
+		get {
+			return XPCServer.ClientRequirement(messageAcceptor:HardenedMessageAcceptor())
+		}
```
jakaplan, line 68: This can be written as just:
`XPCServer.ClientRequirement(messageAcceptor: HardenedMessageAcceptor())`

Both the `get` and `return` can be made implicit.

Sources/SecureXPC/Server/ClientRequirement.swift
```diff
@@ -163,6 +170,34 @@ fileprivate protocol MessageAcceptor {
     func isEqual(to acceptor: MessageAcceptor) -> Bool
 }
 
+fileprivate struct HardenedMessageAcceptor: MessageAcceptor {
```
jakaplan, line 173: Could you spaces instead of tabs for consistency? At least as shown by GitHub, the indentation here noticeably does not match the rest of this file.

Sources/SecureXPC/Server/ClientRequirement.swift
```diff
@@ -163,6 +170,34 @@ fileprivate protocol MessageAcceptor {
...
+		}
+
+		var staticCode: SecStaticCode?
+		var codeSignInformation: CFDictionary?
+
+		guard SecCodeCopyStaticCode(code, [], &staticCode) == errSecSuccess,
+			  let staticCode = staticCode,
+			  SecCodeCopySigningInformation(staticCode, [], &codeSignInformation) == errSecSuccess,
+			  let codeSignInformation = codeSignInformation as NSDictionary?,
+			  let codeFlagsInt = codeSignInformation[kSecCodeInfoFlags] as? NSNumber
```
jakaplan, line 186: Can this directly be cast to `UInt32` here and then later not need to use `codeFlagsInt.uint32Value`?

jakaplan (changes requested):
Thanks for creating this PR!

Just have a little bit of small feedback. Additionally, could you add `hardenedRuntime` to the DocC Requirements section?

jakaplan:
Thanks for contributing!

Issue #128, opened 2023-04-18
SMAppService and MachServiceCriteria


rurza:
Hey! 
First of all, AMAZING work! 
I'm trying to use the framework to talk with a daemon launched by the new `SMAppService` api. The daemon launches, launches server but I can't make it to talk with the client. It fails on the first condition. But...
I'm not sure that `validateThisProcessIsAnSMAppServiceDaemon` should work at all. It checks for the `Bundle`s `bundleURL` and it assumes that it'll receive a path with the `Contents` folder. It doesn't work in my case, it returns a path, but to the bundle itself (so the last component is `*.app`). Am I doing something wrong?

https://github.com/trilemma-dev/SecureXPC/blob/c1050545168792c44984c3cc0ce2fa006c84e188/Sources/SecureXPC/Server/MachServiceCriteria.swift#L561

jakaplan:
Thanks for opening this issue, I'll take a look this weekend.

rurza:
@jakaplan I have a fork that "works". Please take a look :)

jakaplan:
I don't see anything immediately wrong with the code as it currently exists. Could you provide me with the error message that's associated with the `XPCError.misconfiguredServer` error? (If you don't currently have a good way of getting access to this, the easiest is using `NSLog(...)` and looking for the result in Console.app)

---

> It checks for the `Bundle`s `bundleURL` and it assumes that it'll receive a path with the `Contents` folder.

It validates that the daemon or agent is located within its parent's application bundle. This means a `Contents` folder must exist in its path.

> It doesn't work in my case, it returns a path, but to the bundle itself (so the last component is `*.app`).

When you "it returns a path" what is "it" referring to? 

> Am I doing something wrong?

Possibly. Are you creating the `XPCServer` in the daemon? And where is the daemon located within the parent app's directory structure?

jakaplan:
> @jakaplan I have a fork that "works". Please take a look :)

I took a look, but don't understand them. Are you trying to run the `XPCServer` from within the app as opposed to the daemon?

rurza:
@jakaplan 
> It validates that the daemon or agent is located within its parent's application bundle. This means a Contents folder must exist in its path.
> I don't see anything immediately wrong with the code as it currently exists. Could you provide me with the error message that's associated with the XPCError.misconfiguredServer error? (If you don't currently have a good way of getting access to this, the easiest is using NSLog(...) and looking for the result in Console.app)

[...]

PR #131, opened 2023-04-26
Fix parentBundle url method

rurza:
Related to https://github.com/trilemma-dev/SecureXPC/issues/128

### DIFF
--- a/Sources/SecureXPC/Server/ClientRequirement.swift
+++ b/Sources/SecureXPC/Server/ClientRequirement.swift
@@ -142,13 +142,14 @@ extension XPCServer.ClientRequirement {
     
     private static var parentBundleURL: URL {
         get throws {
-            let components = Bundle.main.bundleURL.pathComponents
-            guard let contentsIndex = components.lastIndex(of: "Contents") else {
-                throw XPCError.misconfiguredServer(description: "This server does not have a parent bundle.\n" +
-                                                                "Path components: \(components)")
+            let bundleURL = Bundle.main.bundleURL
+            guard bundleURL.pathComponents.last?.hasSuffix(".app") == true else {
+                throw XPCError.misconfiguredServer(description: """
+                This server does not have a parent bundle.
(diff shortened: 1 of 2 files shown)


Sources/SecureXPC/Server/ClientRequirement.swift
```diff
@@ -142,13 +142,14 @@ extension XPCServer.ClientRequirement {
...
+            let bundleURL = Bundle.main.bundleURL
+            guard bundleURL.pathComponents.last?.hasSuffix(".app") == true else {
+                throw XPCError.misconfiguredServer(description: """
+                This server does not have a parent bundle.
+                Path: \(bundleURL.absoluteString)
+                """)
             }
-            
-            return URL(fileURLWithPath: "/" + components[1..<contentsIndex].joined(separator: "/"))
+            return bundleURL
```
jakaplan, line 152: Why does this return its own bundle URL, but only if it ends with `.app`?

Unless I'm missing something var block will no longer return the parent's bundle URL under any circumstances.

Sources/SecureXPC/Server/MachServiceCriteria.swift
```diff
@@ -499,16 +499,14 @@ private func throwIfSandboxedAndThisLoginItemCannotCommunicateOverXPC() throws {
 // MARK: SMAppService daemon & agent
 
 private func parentAppURL() throws -> URL {
-    let components = Bundle.main.bundleURL.pathComponents
-    guard let contentsIndex = components.lastIndex(of: "Contents"),
-          components[components.index(before: contentsIndex)].hasSuffix(".app") else {
+    let bundleURL = Bundle.main.bundleURL
+    guard bundleURL.pathComponents.last?.hasSuffix(".app") == true else {
```
jakaplan, line 503: Why would the bundle URL of the server be expected to end in `.app`?

jakaplan (changes requested):
I'm unsure what these code changes are attempting to accomplish.

